PR #17, opened 2020-11-25
Title Style

lkiesow:
This patch updates the title which gets a little bit more space while
using the available white-space more effective. This also protects
against overly long title which are ended in a horizontal ellipsis,
ensuring that they never get more than a single line.

---

Preview:
editor-title

### DIFF
--- a/src/main/Video.tsx
+++ b/src/main/Video.tsx
@@ -64,7 +64,7 @@ const Video: React.FC<{}> = () => {
     flexDirection: 'column' as const,
     justifyContent: 'center',
     alignItems: 'center',
-    padding: '10px',
+    padding: '0px',
     borderBottom: '1px solid #BBB',
   });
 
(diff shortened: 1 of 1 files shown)


src/main/Video.tsx
```diff
@@ -247,15 +247,21 @@ const VideoHeader: React.FC<{}> = () => {
...
+  } as React.CSSProperties;
+
 
   return (
-    <div title="Video Area Header">
-      <div css={titleStyle} title="Video Title">{title}</div>
-      <div title="Video Presenters">by {presenters.join(", ")}</div>
+    <div title="Video Area Header" css={{fontSize: '16px'}}>
+      <div style={titleStyle} css={{fontSize: '24px', fontWeight: 'bold'}} title="Video Title">{title}</div>
+      <div style={titleStyle} title="Video Presenters">by {presenters.join(", ")}</div>
```
Arnei, line 262: Instead of using the `style` tag (and splitting the css between the `style` and the `css` tag), I think it would be better to create multiple css objects and apply them using composition:
```suggestion
  const titleStyle = css({
    display: 'inline-block',
    padding: '15px',
    overflow: 'hidden',
    whiteSpace: "nowrap",
    textOverflow: 'ellipsis',
    maxWidth: '500px'
  })

  const titleStyleBold = css({
    fontWeight: 'bold',
    fontSize: '24px',
  })


  return (
    <div title="Video Area Header" css={{fontSize: '16px'}}>
      <div css={[titleStyle, titleStyleBold]} title="Video Title">{title}</div>
      <div css={titleStyle} title="Video Presenters">by {presenters.join(", ")}</div>
    </div>
  );
```
(Only the first three lines are marked as deleted because Github won't let me mark everything in one go)

Arnei (commented):
> ensuring that they never get more than a single line.

If I try really hard, I can get the presenter names to slide below the title. Although having max length on both text elements and windows width below 1300 could be considered an edge case^^

Peek 2020-11-27 11-19

src/main/Video.tsx
lkiesow: Are you certain which font-size will with if we do it like this?

lkiesow:
> >  ensuring that they never get more than a single line.
> 
> If I try really hard, I can get the presenter names to slide below the title. 

Sorry, that wasn't clear. I meant a single line per element. That title and author break into two lines is something I wanted to achieve (imaging a portrait mode smartphone later)

src/main/Video.tsx
Arnei: I am struggling to find concrete proof as to why exactly a childs CSS will always take precedence over a parents CSS, but I don't know why it should not be so.

lkiesow: You are right. I misread the code. Sorry.

Arnei: Do you agree with suggested changes otherwise? Asking in the interest of getting this merged.

lkiesow: Adopted your suggestions

PR #19, opened 2020-11-25
Video Area Code Cleanup

lkiesow:
These patch just removes a few things which aren't actually used or have
no effect at all.

### DIFF
--- a/src/main/Video.tsx
+++ b/src/main/Video.tsx
@@ -7,10 +7,10 @@ import { css } from '@emotion/core'
 
 import { useSelector, useDispatch } from 'react-redux';
 import {
-  selectIsPlaying, selectCurrentlyAt, selectCurrentlyAtInSeconds, setIsPlaying, setCurrentlyAtInSeconds, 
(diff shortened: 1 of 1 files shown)


src/main/Video.tsx
```diff
@@ -131,19 +131,13 @@ const VideoPlayer: React.FC<{url: string, isMuted: boolean}> = ({url, isMuted})
     }
   })
 
-  const playerStyle = css({
```
Arnei, line 134: This has indeed no effect, but the reason for that seems to be the current layout. Would it be wise to keep the minimum size in case the layout changes again in the future?

Arnei (changes requested):
It seems I have a trailing whitespace problem, thanks for pointing it out.

src/main/Video.tsx
lkiesow: I don't think we should keep things in the code for the reason that if we change something else, it might become useful. That just leads to code you don't understand because half of it is in there for no reason. Also, would you really want this to have a minimum size, even if you don't have the space for it? E.g. let's say we have multiple videos later on. That could easily cause issues. Overextending the video over the edges of the display also does not really help.

Arnei: Sound logic, approved.

PR #20, opened 2020-11-25
Timeline Border Radius

lkiesow:
This patch removes the border radius from the timeline since it looks a
bit odd with the scrubber extending over the edge and we have a full
width box with no radius already anyway (the player).


Arnei:
This is just my non-graphic-designer opinion, but I'd rather keep the border radius on the segments. I think it gives them a more "soft" and "friendly" look, while also helping to visually establish a segment as a single entity. I'd agree to significantly reducing the border radius though.

Also the only reason the player does not have a border radius is because I have not found a good way to add one yet :)

PR #26, opened 2020-11-27
Icons for Segment Merge Actions

lkiesow:
This patch adds icons for the “Merge Left” and “Merge Right” segment
actions.

### DIFF
--- a/src/main/CuttingActions.tsx
+++ b/src/main/CuttingActions.tsx
@@ -1,7 +1,15 @@
 import React from "react";
 
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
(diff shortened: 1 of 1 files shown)


Arnei:
Using the `faStepBackward` and `faStepForward` icons may be problematic if we ever get around to implementing actual step video controls like in the old editor (which is very useful according to Rüdiger).

lkiesow:
But that's not on our roadmap. So the question should be, does this icon work for this job. I think it's actually a fairly good depiction of what the actions do.

Besides, you could easily use [angle-right](https://fontawesome.com/icons/angle-right?style=solid) for stepping and [angle-double-right](https://fontawesome.com/icons/angle-double-right?style=solid) for jumping to the next segment.

Issue #22, opened 2020-11-25
Player Max Height


lkiesow:
Setting the `ReactPlayer`'s `width` and `height` attributes causes the player to be adjusted according to the page height, ignoring all limitation for the height (`maxHeight` has no effect). This means that you will  get a huge player stealing all of the space if there is just a single video.

Arnei:
Solved by #111

PR #28, opened 2020-11-27
Introduced a max height for the video player area

Arnei:
Possible solution for #22. This should avoid video elements taking up too much space.


lkiesow:
Sorry, but this does not work. The ReactPlayer's code will still set the size of the video itself which will just cause the complete UI to break. A possible solution would be to sou round the videos with a div and control the size using that. This might mean though that we need to calculate the size ourselves in a multi-video environment

Editor in 4k with video over controls:
editor-4k

lkiesow:
Closing this since the attempt did not work and the code has conflicts by now anyway

PR #52, opened 2020-12-15
Added wrapper around video players to limit video-height

jduehring:
This PR should be a suitable solution for #22. It adds a wrapper around the video player and sets its height to 50% of the current viewport. This should prevent videos from overscaling.

### DIFF
--- a/src/main/Video.tsx
+++ b/src/main/Video.tsx
@@ -134,18 +134,26 @@ const VideoPlayer: React.FC<{url: string, isMuted: boolean}> = ({url, isMuted}) 
     }
   })
 
+  const playerWrapper = css({
+    backgroundColor: 'rgba(255,255,255,1)',
+    width: '100%',
+    height: '50vh'
+  });
+
   return (
-    <ReactPlayer url={url}
-      ref={ref}
-      width='100%'
-      height='auto'
-      playing={isPlaying}
-      muted={isMuted}
-      onProgress={onProgressCallback}
-      progressInterval={100}
-      onReady={onReadyCallback}
-      onEnded={onEndedCallback}
-    />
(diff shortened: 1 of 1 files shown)


src/main/Video.tsx
```diff
@@ -134,18 +134,26 @@ const VideoPlayer: React.FC<{url: string, isMuted: boolean}> = ({url, isMuted})
     }
   })
 
+  const playerWrapper = css({
+    backgroundColor: 'rgba(255,255,255,1)',
```
Arnei, line 138: Probably either want to leave this out (so it defaults to the globally defined backgroundColor), or pick a color with a higher contrast, maybe black. I'd prefer the global default.

Arnei (commented):
Seems to solve the problem.

However also forces the video area to be half of the viewport, which can leave a lot of empty space in certain situations. Currently on the fence about if this is actually a problem or not.

jduehring:
> However also forces the video area to be half of the viewport, which can leave a lot of empty space in certain situations. Currently on the fence about if this is actually a problem or not.

Yeah, I figured that this might not be the optimal solution. I tried different approaches to just limit the height of the player, for example by setting a `maxHeight` in the wrapper. But even when the surrounding div was limited, I couldn't quite get the player to follow these restrictions yet.

src/main/Video.tsx
jduehring: changed

lkiesow:
I agree that it's problematic in some modes:

editor-video-spacing

Can you limit the container height to the height of the videos to not waste so much space? We don't need the waveform to be alligned at the bottom.

Arnei:
This was resolved by the alternative solution in #111, so I'll close this.

PR #62, opened 2020-12-18
Can now drag the scrubber during video playback

### LINKED ISSUE #59: Drag position slider while playing
You cannot drag the position slider while playing which is annoying if you want to e.g. quickly rewind a bit. It would be nice if starting to drag would pause the video but constantly adjust the video playback position and then automatically start plaing again once the slider is released.

Arnei:
Fixes #59.

Dragging the scrubber during playback now pauses the video and then resumes play after the dragging is stopped


lkiesow (changes requested):
Dragging while playing can cause the scrubber to jump around like crazy after release:

editor-use-scrubber-while-playing

Tested in Firefox 84.0

lkiesow (changes requested):
> JSON.parse: unexpected character at line 1 column 1 of the JSON data

Arnei:
Updated the PR, the test deployment should now work as expected @lkiesow

Issue #76, opened 2020-12-18
Provide a config file


Arnei:
Provide a config file that can be changed and read by the application after building.

Suggested file format for the file is TOML.

lkiesow:
+1 for toml or yaml since that is what other components use as well (and it can contain comments)

PR #89, opened 2021-01-11
Added a config file

Arnei:
This PR solves #76 by adding a toml config file. Settings are now parsed in config.ts. Settings are parsed before the website is first rendered.

Retains the ability to use GET-Parameters and default values.
Validates parsed settings.
All the good bits are adapted from https://github.com/elan-ev/opencast-studio/blob/master/src/settings.js.


Arnei:
Added commit to tackle #90.

PR #111, opened 2021-01-19
Attempt at fixing max height issue #2

Arnei:
Another attempt at fixing issue #22. This PR also adds a wrapper to the video players, but aims at abusing the css attributes `position` and `paddingTop` to make them more responsive. This also forces us to calculate a total aspect ratio for the wrappers, making this solution more complex than I'd personally like it to be.


lkiesow (approved):
This is a big improvement. I still found a few edge cases where we could do better but overall, this works. Merging…

---

Edge case:
Screenshot from 2021-01-22 18-27-55

lkiesow:
Or, well… not merging since there are conflicts. But go ahead and do that once you got them fixed :D

Issue #58, opened 2020-12-17
Jump to position


lkiesow:
Instead of dragging the position indicator, it would be nice if a user could click on the waveform to jump to a certain position in the video.

PR #113, opened 2021-01-19
Can now move the "current point in time" by clicking on the timeline

Arnei:
By clicking on any part of the timeline, the "current point in time" now moves to the clicked point, which should move the scrubber and update the videos. Solves #58 

Maybe TODO: Block mouse click events when the scrubber is clicked. Not sure if we would actually want that or not.


lkiesow:
Unfortunately, this only works if the video is not playing in which case the scrubber jumps to the position to then immediately jump back:


https://user-images.githubusercontent.com/1008395/111482020-cc8ff380-8733-11eb-884e-96d92c695688.mp4

PR #127, opened 2021-01-22
Shorten and align save and discard

lkiesow:
This patch shortens the texts for saving and discarding the current
project and add a header alike the one you get when choosing to start
processing.

### DIFF
--- a/src/main/Save.tsx
+++ b/src/main/Save.tsx
@@ -36,10 +36,10 @@ const Save : React.FC<{}> = () => {
 
   return (
     <div css={saveStyle} title="Save Area">
-      <span>
-        Here you can save the changes you made, but the video will not be cut yet. <br />
-        To make Opencast cut the video, please go back and select "Start processing". <br />
-        Do you truly wish to save?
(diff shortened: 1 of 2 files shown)


src/main/Save.tsx
```diff
@@ -36,10 +36,10 @@ const Save : React.FC<{}> = () => {
...
   return (
     <div css={saveStyle} title="Save Area">
-      <span>
-        Here you can save the changes you made, but the video will not be cut yet. <br />
-        To make Opencast cut the video, please go back and select "Start processing". <br />
-        Do you truly wish to save?
+      <h1>Save current project</h1>
+      <span css={{maxWidth: '500px'}}>
+        The video will not be processed but all cutting information are stored
+        in Opencast. You can continue your eddi later.
```
Arnei, line 42:
```suggestion
        The video will not be processed but all cutting information will be stored
        in Opencast. You can continue your edit later.
```

Arnei (changes requested):
Seems good to me, particularly the consistency this introduces between the different finish options.

PR #131, opened 2021-01-22
Workflow Selection View Improvements

lkiesow:
This patch slightly improves the workflow selection view:

- We shouldn't recommend workflows which might not exist due to an
  administrators chiuce and should hope that admins properly lable them
  instead.
- This adds some padding to make this look better an smaller screens
- We shouldn't show a development(?) comment when there is no workflow
  description.

### DIFF
--- a/src/main/WorkflowSelection.tsx
+++ b/src/main/WorkflowSelection.tsx
@@ -49,13 +49,12 @@ const WorkflowSelection : React.FC<{}> = () => {
         workflows[selectedWorkflowIndex].description
       );
     } else {
-      return (
-        "And this is where I would put a workflow description.... if I had one!"
-      );
+      return (' ');
     }
   }
 
   const workflowSelectionStyle = css({
+    padding: '20px',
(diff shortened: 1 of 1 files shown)


src/main/WorkflowSelection.tsx
```diff
@@ -49,13 +49,12 @@ const WorkflowSelection : React.FC<{}> = () => {
         workflows[selectedWorkflowIndex].description
       );
     } else {
-      return (
-        "And this is where I would put a workflow description.... if I had one!"
-      );
+      return (' ');
```
Arnei, line 52: Would it be better to show nothing or _"No workflow description available"_.

Arnei (approved):

src/main/WorkflowSelection.tsx
lkiesow: I don't think “No workflow description available” is helpful for users.
They cannot do anything about it and it doesn't help them to know what the workflow is about.
It's just distracting from other elements.

Issue #129, opened 2021-01-22
Vertically align “Finish” selection if there is not enough space


lkiesow:
The view “Finish” should vertically align the selection if there is not enough horizontal space.

---

![Screenshot from 2021-01-22 18-12-21](https://user-images.githubusercontent.com/1008395/105522475-9444d980-5cdd-11eb-8972-6f4a96cef987.png)

---

PR #146, opened 2021-01-25
Introduced wrapping to the finish menu

Arnei:
Allows flexWrap in the finish menu ("Save, Process, Discard") to avoid squishing the large buttons on smaller resolutions.

This is the easiest solution to #129, but may not look as good, since the buttons do not immediately change from row to column. Another solution would be media-queries, but those are require specifying a width and are thus not quite as dynamic.


lkiesow:
Definitely an improvement → merging

PR #154, opened 2021-01-26
First basic unit test for reducers of videoSlice

Arnei:
Instead of spouting error messages, `npm test` should now actually run tests. To commemorate this, this PR includes some unit tests for the reducers of `videoSlice`.


lkiesow:
This breaks the video slider. See #164

PR #160, opened 2021-01-29
Updated the workflow selection rendering process

### LINKED ISSUE #128: Skip selection of a single workflow
In most cases, it's best to not let users select different  workflows but select one they should use for processing. If there is just one workflow we should completely hide the workflow selection instead of presenting a selection with a single workflow.

Related code:

https://github.com/elan-ev/opencast-editor/blob/35e3f668c4b3d26873b3dde228156fd563e74ada/src/main/WorkflowSelection.tsx#L32-L38

Arnei:
- Now displays a more proper version of the page when there are no workflows available
- Added a new version of the page for when there is exactly one workflow available

Somewhat resolves #128

### DIFF
--- a/src/main/Save.tsx
+++ b/src/main/Save.tsx
@@ -56,7 +56,7 @@ const Save : React.FC<{}> = () => {
 /**
  * Button that sends a post request to save current changes
  */
-const SaveButton: React.FC<{}> = () => {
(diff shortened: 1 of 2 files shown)


lkiesow (approved):
Looks good. Has conflicts.

PR #175, opened 2021-02-04
Start Processing Label

lkiesow:
This patch changes the label “Start processing with workflow” to simply
read “Start processing” since the “with workflow” does not actually add
any valuable information here but may lead to some confusion since it
(kind of) reveals internal workings of Opencast.


Arnei:
Closing because I unfortunately broke this and making a new PR #179 is faster than fixing this up.

Issue #190, opened 2021-02-18
Introduce Internationalization


lkiesow:
At the moment, all strings in the editor are in English. It would be great if we could add a language selector (maybe just select the browser language) and translate the tool into several languages like we did e.g. with [Opencast Studio](https://github.com/elan-ev/opencast-studio) where we used `i18next` for this.

Arnei:
Partially resolved by #205, which offers translation  and automatic language selection based on browser language, but no language selector. In my book this is good enough to close this issue. If we still want a language selector, we can always create a new issue.

PR #205, opened 2021-03-03
Introducing Internationalization

jduehring:
This PR is introducing Internationalization to the editor using i18next, as proposed in #190. All strings have been changed to translation-variables which will use a translation-string corresponding to the selected language. Currently the only way to select different languages is by changing your browser language. 
New language support can easily be achieved by creating a corresponding json file, translating all necessary strings and adding the file to the `config.tsx`.
A few german translations are already added.


Arnei (changes requested):
Works and looks good overall.

I would probably refrain from including the german translation file. While it makes for a good example that i18n is actually working, a half-complete translation is likely to be more confusing than helpful to users. 

This is more of a style question, but what made you choose
``` 
    "mainMenu-cutting-button": "Cutting",
    "mainMenu-finish-button": "Fertigstellen",
```
over
```
    "mainMenu": {
      "cutting-button": "Cutting",
      "finish-button": "Fertigstellen",
    }
```


Independently, grouping by component seems fine given the project structure.

jduehring:
> This is more of a style question, but what made you choose
> 
> ```
>     "mainMenu-cutting-button": "Cutting",
>     "mainMenu-finish-button": "Fertigstellen",
> ```
> 
> over
> 
> ```
>     "mainMenu": {
>       "cutting-button": "Cutting",
>       "finish-button": "Fertigstellen",
>     }
> ```

Committed all the requested changes. 
Regarding the style of the json files, I honestly considered the more structured approach, but at that point I was already pretty far into building the file. I also thought just for translating it might be a bit easier to overlook the whole file. But I guess at the end it depends on personal taste.

Arnei:
> Regarding the style of the json files, I honestly considered the more structured approach, but at that point I was already pretty far into building the file. I also thought just for translating it might be a bit easier to overlook the whole file. But I guess at the end it depends on personal taste.

I asked around the team a bit, and it seems that not nesting translation files is something that leads to regrets later on.  I also don't think the file is particularly easy to overlook in its current state. Therefore I would be of the opinion to rework the translation file structure, even though that means a lot of string replacement again. 

And while we're on the topic, YAML looks nicer to me than JSON (example: https://github.com/elan-ev/tobira/blob/master/frontend/src/i18n/locales/en.yaml). But that might just be my inner python syntax lover speaking.

Arnei (approved):

PR #214, opened 2021-03-10
Metadata

### LINKED ISSUE #199: Allow editing DC metadata
We want to be able to edit basic Dublin Core metadata fields like title, presenter, language or description.

Arnei:
This PR allows the user to edit metadata of an event. Resolves #199

- Enables a main menu item for metadata.
- Gets metadata from Opencast
- Displays Metadata in a form that allows for changes
- Sends metadata to Opencast

TODOs:
- Start "Republish metadata" in Opencast if necessary?
- Get rid of error message in console
- Automatic validation of date type form entries


lkiesow:
- Do we want fields like “Presenter” to be a drop-down? Or a simple text field? I think the latter may make more sense and cause less problems.
- How do we handle fields with multiple entries (again thinking of presenter)

Arnei:
>     * Do we want fields like “Presenter” to be a drop-down? Or a simple text field? I think the latter may make more sense and cause less problems.
> 

Ideally we probably want it to be both a drop-down and an input field, similar to the Admin-UI. I personally do like the quasi-autocomplete functionality this leads to. But might be more error prone.

>     * How do we handle fields with multiple entries (again thinking of presenter)

Hoping no one ever tries to use multiple entries is not an option, is it? In that case, we again probably want it work similarly to how it currently works in the Admin-UI (Display multiple values, allow removing exisiting values, onEdit also show an input/select element to add new elements)

lkiesow:
> Make changes as you like, then hit the Submit button.
> Note that you will still have to start processing for your changes to take effect.

The first like is pretty meaningless and from what I can tell, they take effect immediately. Just not for publications.

> Note that updated metadata are not immediately published.
> You need to start a workflow to update publications.

Also, I would probably call the button „Save“ and the title „Save your changes“.

---

You also lose some metadata when switching between views:


https://user-images.githubusercontent.com/1008395/111484344-f813dd80-8735-11eb-829b-d4563fcbe940.mp4

Finally, I'm not sure where the „metadata.Arnes Metadaten“ come from. Are these already extended metadata?

Arnei:
> The first like is pretty meaningless and from what I can tell, they take effect immediately. Just not for publications.

Yes, changes are immediately saved to Opencast when hitting "Submit". But users may not understand the difference between "saved to Opencast" and "saved to publications". So I wanted to give users a hint that even though they hit "Submit" (or "Save"), they would still need to select a processing step for their changes to become effective in their Moodle-Video or whatever. Seems like I should mull some more over the wording.

> Also, I would probably call the button „Save“ and the title „Save your changes“.

Left this as "Submit" to avoid confusion with the "Save" button in the "Finish" menu (since they do different things).

> You also lose some metadata when switching between views:

That is weird, good catch.

> Finally, I'm not sure where the „metadata.Arnes Metadaten“ come from. Are these already extended metadata?

Yup, that's already an extended metadata catalog. Excuse me restarting pyca.opencast.org a couple of times during your testing.

lkiesow:
> Left this as "Submit" to avoid confusion with the "Save" button in the "Finish" menu (since they do different things).

Don't both save data to the asset manager?
But maybe a bit more explicit: “Save metadata”

Arnei:
> Or maybe even remove the submit button and just use the save button in the finish view.
> That way, you also have no conflicts with the “Discard changes” functionality which will not discard these changes at all.

I went ahead and implemented it this way. This is closest to how it works in the Admin-UI as well.

PR #220, opened 2021-03-12
Added hotkey support for buttons in the cutting view

### LINKED ISSUE #75: Keyboard shortcuts
Improve usability and efficiency of long term users through keyboard shortcuts.

Shortcuts for:
- All cutting actions
- Video controls (play/pause)
- Menu switching
- Quick finish (Completely bypass the finish menu by automatically start processing in Opencast with the default workflow)
- Other

Arnei:
This PR introduces hotkey support for buttons in the cutting view through react-hotkeys. Resolves #75

Also reworked the existing hotkeys for the scrubber to fit with the react-hotkeys schema.

Since it is amazingly difficult to find key combinations that aren't already used by SOMETHING, any feedback on possible key values for the currently very simple ones would be appreciated.


lkiesow:
Works well except for space:

1. Click on the cut (or really any) button with your mouse
2. Hit space
3. The button is triggered
4. Play/Pause is not triggered

This is especially annoying if one of the merge buttons is the last button you clicked on.

We probably also want a view showing the available keyboard bindings

Arnei:
> Works well except for space:
> 
>     1. Click on the cut (or really any) button with your mouse
> 
>     2. Hit space
> 
>     3. The button is triggered
> 
>     4. Play/Pause is not triggered
> 
> 
> This is especially annoying if one of the merge buttons is the last button you clicked on.

Thanks for catching that. Likely has to do something with keyboard controls. If you navigate the page by keyboard, you should be able to activate button-type UI with space-bar, and I wanted to avoid playing/pausing the video everytime someone just wanted to push a button. I'll see if I can figure something out.

> We probably also want a view showing the available keyboard bindings

If we ever get around to that header we talked about, that would probably be the perfect place to put a hotkey overview. With the current UI, I've got no good idea on how to fit it in.

lkiesow:
> If we ever get around to that header we talked about, that would probably be the perfect place to put a hotkey overview. With the current UI, I've got no good idea on how to fit it in.

That's exactly what I thought as well.

lkiesow:
I still feel like the current space behavior is more harmful than helpful.
Using space for a video player is very common and currently, you may destroy parts of your work if you use that.

Maybe, if we want to preserve keyboard navigation and support `Spacebar` for button-selection in addition to `Enter`, we could add something like `element.blur()` to the mouseup event listener. This would ensure that you don't unknowingly focus an element by clicking on it to use the functionality. But if you use the keyboard to navigate to the button, you will stay there and will be able to use space to trigger that button's function. Something like:

```html
<button onclick="console.info('click')"
        onmouseup="console.info('mouseup'); this.blur()">
    test
</button>
```

Arnei:
> Maybe, if we want to preserve keyboard navigation and support `Spacebar` for button-selection in addition to `Enter`, we could add something like `element.blur()` to the mouseup event listener. 

There seems to be no standard way for doing this, so thanks for the suggestion. I went ahead and implemented it.

PR #255, opened 2021-04-12
Duration fallback

### LINKED ISSUE #253: Broken scrubber if there are no segments
Testing on develop.opencast.org I found a broken editor:

editor-broken-scrubber-no-segments

My guess is that the problem might be related to the event having no segments in which case I would expect the editor to assume a single segment for the recording.

Arnei:
The duration return by the backend can be "null". This is not handled by the frontend at all, resulting in an erroneous state.

Now, the frontend will try to determine the duration from the primary video player if there still is no duration by the time the video player is done loading.

Resolves #253.

DO NOT MERGE, this relies on a fix in the backend by @lwillmann that is not yet implemented.


Arnei:
Tested this some more:
* Determining duration fails for both of my test cases. The determined duration is "Infinity".
* Even if the duration determination somehow works, the video will not be processed (editor operation is skipped).

Arnei:
Closing this because this solution has found no appeal.

PR #308, opened 2021-05-20
Added error page

### LINKED ISSUE #162: Provide Error Page
If an incorrect media package identifier is provided, the editor interface mostly breaks. 
It would be great to have a proper error page instead, possibly even displaying the error type if it is known based on the `editor.json` response.

Screenshot from 2021-02-01 18-14-43

Arnei:
Introduces an error page that can be displayed in case of an unrecoverable error.
Failures that occur when fetching information for Opencast will now cause the error page to appear.

Can be tested by calling the editor with a non-existant mediaPackageId, for example `?mediaPackageId=nonsense`.

Partially resolves #162 (Error page is not yet shown if the media package exists but the contents are faulty).

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -161,6 +161,11 @@
       }
     },
 
+    "error": {
+      "generic-message": "A critical error has occurred!",
+      "details": "Details: "
+    },
+
     "various": {
       "error-details-text": "Details: {{errorMessage}}\n",
       "error-noDetails-text": "No error details are available.",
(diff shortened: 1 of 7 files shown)


lkiesow:
- Why  have a fixed min size on this page (in preparation for making this mobile-friendly eventually). There is really no reason here
- Breaking lines overlap and I see no reason for setting line-height

Screenshot from 2021-05-20 17-20-01

lkiesow (approved):

PR #311, opened 2021-05-21
Added a validator for datetime metadata fields

### LINKED ISSUE #297: Time- and Datepicker do not validate entered values
Date can be added with pretty incorrect formatting (see 1 and 2) and will still be accepted
        1. Add line of text, no error
        2. Replace numbers (at any point of date or time) with one letter, no error
-Sonia

Arnei:
For whatever reason, `mui-rff` disables the automatic datetime validation from material-ui, so we have to do the validation ourselves.

Resolves #297

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11764,9 +11764,9 @@
       "integrity": "sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w=="
     },
     "mui-rff": {
-      "version": "3.0.6",
-      "resolved": "https://registry.npmjs.org/mui-rff/-/mui-rff-3.0.6.tgz",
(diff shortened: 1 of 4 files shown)


lkiesow:
Non-set dates are now also marked as invalid:


https://user-images.githubusercontent.com/1008395/119152684-d47c5480-ba50-11eb-8f5c-d6480f4e6392.mp4

lkiesow:
Typing in a time, the field is also marked as invalid:

Screenshot from 2021-05-21 16-24-49

Arnei:
> Typing in a time, the field is also marked as invalid:
> 

Curiously, when I enter "00:00" as a time it is not marked as invalid.

lkiesow:
There still seem to be several cases where this breaks:


https://user-images.githubusercontent.com/1008395/119548034-15989f80-bd96-11eb-883c-1f3d7a1703fc.mp4

lkiesow:
One Problem persists:
Screenshot from 2021-05-31 12-44-35

Arnei:
The underlying issue here seems to be that the KeyboardDateTimePicker does not seem to be able to handle the empty string correctly (which is how Opencast represent "no date"). Should be fixed now.

PR #315, opened 2021-05-25
Added a config option for displaying metadata fields

### LINKED ISSUE #38: Make Metadata Fields Configurable
To not confuse users, it would be nice to reduce the set of available metadata fields to what is actually being used at the institution. For this, we should introduce a setting which allows to specify a list of fields to show as sub-set of all available fields. If no list is provided, all fields are displayed.

Arnei:
Added an option to the editor-settings that lets users control which metadata catalogs and fields are displayed in the editor.

Resolves #38. Does not resolve the follow up request regarding read-only configuration in the same issue.

### DIFF
--- a/public/editor-settings.toml
+++ b/public/editor-settings.toml
@@ -20,6 +20,23 @@ password = "opencast"
 [metadata]
 # If false, the metadata button is not displayed in the main menu
 show = true
+## Configure metadata display
+## Here you can set which metadata fields should be displayed in the editor
+## for each catalog.
+## If the catalog is not specified, all fields of that catalog will be
+## displayed.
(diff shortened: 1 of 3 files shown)


lkiesow (approved):
Works as expected, though I have a suggestion for a slight configuration change __if__ it's easy to change. If not, lets keep this.

While I can hide catalogs with the current mechanism, I have to remember to do that when I change the admin interface configuration and add a new catalog there. A safer way would be to say:

- Hide all catalogs which are not configured if any metadata configuration is present
- Add a simple configuration to show the whole catalog. E.g. something like:

          "Arne's Metadata = *

lkiesow:
No time left. Let's merge it like this for now and I'll add my suggestions as a new issue.

PR #313, opened 2021-05-25
Change scrubber handle design

Arnei:
Moves the scrubber head to the top of the timeline, to make the waveform more visible.
Also changes the design to be more in line with traditional video editors.

Resolves  #153

PR #316, opened 2021-05-26
Stopped vertical movement of scrubber

Arnei:
The scrubber would sometimes move vertically when dragging. This enforces setting y to 0.

Mentioned in #313.


lkiesow:
I cannot reproduce the problem with the old scrubber but this still workes and I assume it will fix the problem with the new one. Merging.

PR #321, opened 2021-05-27
Reworked errorBox messages

### LINKED ISSUE #171: Editor “unusable” if event is being processed
If an event is already being processed, there are no helpful messages in the editor. Users will just end up with an error when they try to start processing the event:

Screenshot from 2021-02-03 17-02-22

### LINKED ISSUE #269: "Internal Server Error" and other unhelpful error messages.
Clicking on "Save changes" or "Start processing" can lead to error messages like "Internal Server Error" or others. These are not helpful to the user, as they are left clueless on what the problem is, how to fix it or what to do next. -Wien

Arnei:
Reworked errorBoxes (for when you do not want display an error page).
ErrorBoxes should now be more consistent, more informative and easier to integrate into new components.

Resolves #269, Resolves #171.

Bildschirmfoto vom 2021-05-27 16-38-53


lkiesow:
- Weird mixture of alignments (partly aligned left, partly centered). Pick one
- Extremely wordy with confusing error code. What about shrinking the whole thing down to:
  - _**Error:** Opencast is already running a workflow on this event. You have to wait until it finishes before you can perform this action._

Arnei:
> Extremely wordy with confusing error code. 
>       

Seems I got too enthusiastic about giant error boxes, thanks. I shrank the contents according to your example. Although I kept an additional sentence that specifies which action the error pertains to.

lkiesow:
I still think this is rather lengthy. Any reason for going with this long variant? Also, the current wording suggests that Opencast cannot run any workflows in parallel. And the “Details: Locked” is somewhat confusing for end-users, isn't it?

---

current state:

Screenshot from 2021-06-07 14-36-10

---

suggestion:

Screenshot from 2021-06-07 14-42-51

naweidner:
A general question: Why is the border of the error box actually dashed?
Wouldn't it maybe look better if we gave the box a background color (like in OC Studio) instead of a dashed border?

If the error message is quite long, the box takes up almost the entire width of the page, couldn't you give the box a maximum width here? (Probably doesn't matter anymore with the shorter error messages I guess?)

Arnei:
> A general question: Why is the border of the error box actually dashed? Wouldn't it maybe look better if we gave the box a background color (like in OC Studio) instead of a dashed border?
> 
> If the error message is quite long, the box takes up almost the entire width of the page, couldn't you give the box a maximum width here? (Probably doesn't matter anymore with the shorter error messages I guess?)

It is dashed because I thought stripes convey danger and thus draw attention (in addition to the red color). But I'm in no way partial to it. Got a picture of how that looks like OC Studio?

So you are suggesting a smaller width would look better? Or what would be the rationale here?

naweidner:
This is what it looks like in OC Studio, the color is maybe a bit too bright (and eye-hurting). This box definitely catches my eye (the dashed border does too, but not as much as the box in the picture).

grafik

> So you are suggesting a smaller width would look better? Or what would be the rationale here?

Yes, maybe it would look better and fits with the rest of the page? But it's just a suggestion. Maybe I'm wrong and it doesn't look good or doesn't draw the attention that much anymore.

Arnei:
Closing this PR by virtue of being heavily outdated.

PR #329, opened 2021-05-31
Display critical error if Opencast does not provide a video duration

### LINKED ISSUE #253: Broken scrubber if there are no segments

Arnei:
The duration return by the backend can be "null". This is not handled by the frontend at all, resulting in an erroneous state.

Now, the critical error page is shown to inform the user that editing is impossible.

Resolves #253 and provides an alternative to #255, as determining the duration in the frontend does not seem to work properly.


lkiesow:
A critical error for a missing duration seems excessive. What exactly do we need this duration for so desperately?

Arnei:
The cutting view relies on the duration to determine where the user is setting cut marks in the cutting view, since the backend asks for absolute timestamps.

It may be fine to just display an error in the cutting view, since this should not effect other views like metadata.

mtneug:
@Arnei did you find out why Opencast does not report durations at times?

Arnei:
> @Arnei did you find out why Opencast does not report durations at times?

I'll have to admit that I do not remember if we ever found out or not.

src/main/Video.tsx
```diff
@@ -58,8 +59,12 @@ const Video: React.FC<{}> = () => {
       } else {
         dispatch(setError({error: true, errorTitle: undefined, errorMessage: t("video.comError-text"), errorDetails: error, errorIcon: undefined}))
       }
+    } else if (videoURLStatus === 'success') {
+      if (!duration) {
```
LukasKalbertodt, line 63: This branch will also be taken for `duration === 0`. While yes, videos with such a duration are not useful anyway, this should likely be `if (duration == null)`. However, TypeScript will complain about that as the type of `duration` is `number` and not `number | null`. This should be fixed (in `videoSlice.ts` I think) so that the type system reflects reality.

src/main/Video.tsx
```diff
@@ -58,8 +59,12 @@ const Video: React.FC<{}> = () => {
...
+        dispatch(setError({error: true, errorTitle: undefined, errorMessage: t("durationError-text"), errorDetails: error, errorIcon: undefined}))
```
LukasKalbertodt, line 64: Just an unrelated note: In the `interface error`, you could change `errorTitle: string | undefined` to `errorTitle?: string`. And then you don't have to write out `errorTitle: undefined` in objects you create, you can just omit that part then. Same goes for the two other potentially undefined fields.

LukasKalbertodt (changes requested):
I have not yet found a way to test this (to reproduce the null duration), but the code is straight forward and I'm pretty sure it does what the PR says. I only found one thing that could be improved.

LukasKalbertodt (approved):
My comments are resolved!

Unfortunately, there is still a type error (see CI)

LukasKalbertodt:
> Maybe that is good indicator for why we shouldn't allow duration to be of type `null`. The editor expects there to be a duration. it is not equipped to work without a duration. So it needs to be provided a duration. if it isn't, it breaks (and now hopefully displays an error message). It does not make much sense to me to add `| null` to every parameter passed by Opencast and add a huge amount of null checks to the code if the code isn't supposed to work with a null value in the first place.

Fair, and I strongly agree with not putting null checks everywhere. But I think the fault lies in the structure here. I think the error should not be "thrown" there but rather in line 200 (`state.duration = action.payload.duration`). Here, the right hand side is `any`. Here, it should be checked whether the value is actually a non-null `number`. If so, assign it to `state.duration` (which has the type `number` without null). If not, throw an error. But already letting a `null` value INTO `state.duration` and only catching it later in a component seems suboptimal to me. Also, who says that that `Video` component is the first to receive that `null` duration? Maybe in the future, it just so happens that another component receives it and already breaks. 
I have too little knowledge of the redux structure and stuff, but maybe actually `throw`ing an error in that line and using error boundaries is a better solution?

To be clear: I don't want to waltz in here, into a project I haven't written code for, and tell yall how to run your business :D Just some information and opinions. Do with that what you will ^_^

Arnei:
Do agree that the error should be thrown asap in the line you suggested. Unfortunately that's non-trivial (or at least way more difficult than throwing the error `Video.tsx`), so I didn't do that.

LukasKalbertodt (approved):

PR #346, opened 2021-06-01
Fixed waveform generation always throwing error

Arnei:
The error callback for the waveform generation was not working as a proper callback, but would always throw an error. By coincidence, this was hardly noticable on shorter videos, but very noticable on longer ones.

This fixes the callback. "Generating waveform" should now be displayed up until the waveform shows up, even on longer videos. In case of an error (e.g. when the video has no audio), "Waveform could not be generated" will still show.


lkiesow:
Seems to work → merging

PR #348, opened 2021-06-03
Display error if a workflow is running on the event

Arnei:
Uses new information from the `edit.json` endpoint to stop users from making changes that could not currently be saved or processed.

Related to #171.


lkiesow:
Let's wait for https://github.com/opencast/opencast/pull/2675 to be merged

src/redux/videoSlice.ts
```diff
@@ -174,6 +174,10 @@ export const videoSlice = createSlice({
         // });
 
         // New API
+        if (action.payload.workflow_active) {
+          state.status = 'failed'
+          state.error = "An Opencast workflow is currently running, please wait until it is finished."
```
lkiesow, line 179: Internationalization is missing for this.

lkiesow (changes requested):
This looks like the tool or Opencast broke (we don't want to give users that impression) while it should notify the user that things are being processed and he just has to wait:

Screenshot from 2021-08-17 12-56-26

What about [using a gears icon instead](https://fontawesome.com/v5.15/icons/cogs?style=solid) to indicate that work is in progress, and then just display the newly added (and then properly internationalized) error message?

Arnei:
The page now looks like this:
Bildschirmfoto vom 2022-04-21 09-26-25

Issue #362, opened 2021-06-08
Improve metadata fields configurability


ziegenberg:
I should be configurable, which metadata fields are just displayed as read-only vs which can be edited. For example, some fields might add valuable information for context and editing other fields, but they should not be edited in a certain context.

This is a follow-up to #38

lkiesow:
Related to #360

PR #396, opened 2021-08-06
Added readonly as metadata config option

Arnei:
Each field in a metadata catalog can now be set to be editable by the user or not.

Furthermore, this reworks how settings are written in the editor-settings.toml to allow specifiying multiple attributes for one metadata field. If an attribute is not specified, a default is assumed instead.

To test this, play around with example definitions in the editor-settings.toml and observe how they change the metadata ui.


ziegenberg:
Theoretically, this should solve #362, shouldn't it?

lkiesow:
It should. Closing that one.

PR #411, opened 2021-08-19
Switch to `id` parameter

lkiesow:
This patch switches from `mediaPackageId` to `id` as parameter for
providing the media package identifier to work on. The old parameter is
still available as fallback though.


src/config.ts
```diff
@@ -97,6 +97,12 @@ export const init = async () => {
     // Create empty objects for full path (if the key contains '.') and set
     // the value at the end.
     let obj : {[k: string]: any} = rawUrlSettings;
+
+    // Fallback for old parameter
+    if (key === 'mediaPackageId') {
+      key = 'id';
+    }
```
Arnei, line 104:
```suggestion
    // Fallback for old parameter
    if (key === 'mediaPackageId') {
      key = 'id';
    }
    
    
    if (key.startsWith('opencast.')) {
      return;
    }

```

Arnei (approved):
Looks good, approved. The change request below is only to fix the merge conflict.

PR #432, opened 2021-09-03
Fixed videos not loading sometimes when page wasn't cached

### LINKED ISSUE #405: Editor fails to load video
When first opened, a lot of times, the video in the editor will not be loaded. Coming back later or pressing F5 to reload the page will almost always fix the problem. This will often appear when looking at the test deployments. Probably because some things aren't cached, which may cause a timing issue.

---

Hit `Ctrl + F5` a number of times in a private window to get this screenshot:

Screenshot from 2021-08-17 15-24-30

Arnei:
This was caused by the video components setting the dimension of their video as w:0,h:0, causing them to be rendered with no size. Now we instead wait for the resolution to not be 0.

Resolves #405

### DIFF
--- a/src/main/Video.tsx
+++ b/src/main/Video.tsx
@@ -28,6 +28,8 @@ import { useTranslation } from 'react-i18next';
 import { selectTitleFromEpisodeDc } from "../redux/metadataSlice";
 import { setError } from "../redux/errorSlice";
 
+import { sleep } from './../util/utilityFunctions'
+
 /**
  * Container for the videos and their controls
  * TODO: Move fetching to a more central part of the app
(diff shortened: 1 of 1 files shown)


src/main/Video.tsx
```diff
@@ -132,17 +134,29 @@ const VideoPlayer: React.FC<{dataKey: number, url: string, isPrimary: boolean}>
...
-    setReady(true);
-
+  // Tries to get video dimensions from the HTML5 elements until they are not 0,
+  // then updates the store
+  async function updateAspectRatio() {
     if (ref.current && ref.current.getInternalPlayer()) {
       let w = (ref.current.getInternalPlayer() as HTMLVideoElement).videoWidth
       let h = (ref.current.getInternalPlayer() as HTMLVideoElement).videoHeight
+      while (w === 0 || h === 0) {
+        await sleep(1000);
```
lkiesow, line 144: Would it make sense to make this only a tenth of a second  to be more responsive at startup?

```suggestion
        await sleep(100);
```

PR #434, opened 2021-09-08
Send POST request sequentially

### LINKED ISSUE #353: Finish - save changes does not work
Finish - save changes only saves the metadata and not the cutting information. Tested with safari and google chrome.

### LINKED ISSUE #354: Finish - start processing does not save metadata changes (safari)
If i start processing in safari i loose metadata changes.

### LINKED ISSUE #363: Finish - start processing loses the cuts
Most times it works properly. But sometimes the video is published in the original length without any cuts.

Arnei:
The multiple post requests that are send when saving or starting processing are now sent sequentially. Requests wait for their previous requests to return with a successful response code.

This change is supposed to help avoid race conditions than seemingly occur when sending post requests to the metadata and editor endpoints in Opencast at the same time.  Therefore this hopefully resolves #353 resolves #354 resolves #363. It might be worth to look at why exactly these errors occur in the backend.

### DIFF
--- a/src/main/Save.tsx
+++ b/src/main/Save.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 
 import { css } from '@emotion/react'
 import { basicButtonStyle, backOrContinueStyle, ariaLive, errorBoxStyle,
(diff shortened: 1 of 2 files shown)


lkiesow:
Seems reasonable, but I'm wondering if it wouldn't make more sense for the final REST call to require all fields, including metadata so that we do not need two network calls in the first place?

lkiesow:
I've created #485 to not forget this and possibly pick it up in the future. For now, let's get this in.

PR #436, opened 2021-09-14
Added overview page for keyboard controls

### LINKED ISSUE #310: [Feature] Provide an overview of the available keyboard commands
**Problem**

The currently available keyboard commands are only visible if you take your time, hover over the respective buttons and actionable items, and wait for the tooltip. There are several problems:

- If you do not know which elements can be operated with the keyboard, there is no way of discovering the keyboard shortcuts
- There is no central place where all the keyboard shortcuts are listed as a reference
- People not knowing there are keyboard shortcuts most probably won't find them

**Proposed Solution**

There should be a central place for easy access with all available keyboard shortcuts. Eather as an additional menu item (maybe with a question mark) or as a modal popover.
This could also be the place for a very short help text with the basic information on how to use the tool.

I like the way how it's done in Confluence: 
Screenshot from 2021-05-21 09-55-07

Arnei:
Added a button to the far bottom of the main menu that opens a page displaying the key combinations for each of the available hotkeys. If/when hotkeys are added/changed/removed from the editor in future iterations, the overview page will automatically adapt.

Also rewrote hotkey related stuff in other places because I had to change the hotkey keyMap to an object.

Intended to resolve #310.


lkiesow:
> Regarding 1: But not enough to fit the word "video" it seems. I don't see a better way to avoid this issue unfortunately.

Looking at my screenshot: I see more than enough space if there is a break between the two tables, which there is in the screenshot anyway, since the second table does not fit properly.

>  Regarding 2&3: Definitely seems to be caused by the main menu ending prematurely. Do you have any meany by which I can reproduce this?

The screenshot is just from a browser snapped to the left half of my screen. But this also appears in mobile mode or basically any other situation where the content is longer than the menu.

>  Regarding 5: Attempted to move to a centered layout with the recent commits. Will cause overlap with the main menu for small widths, not sure why though.

Yep, that seems to break things:

Screenshot from 2021-10-31 14-09-31

PR #437, opened 2021-09-17
Basic example for how a zoom of the timeline could be implemented

### LINKED ISSUE #107: Make the timeline "zoomable"
Currently the UI doesn't do too well with handling multiple smaller cuts in a row. Especially on video files with long lengths, this can become a problem. To allow for precise graphical cutting, it would be nice to allow users to "zoom into" the timeline. Typically this would mean looking at only a portion of the timeline, in which the scrubber can then be moved with higher precision than normal.

Arnei:
Use the buttons over the top right of the timeline to zoom in and out. Any feedback, particularly on UI design would be appreciated.

Zoomlevel is currently hardcoded between x1 and x2, so no zooming beyond that.

This is still missing translations.

Resolves #107.


Arnei:
Fairly outdated and overall too convoluted. Superseeded by #1033.

PR #441, opened 2021-09-21
Display flatline in case of waveform generation error

### LINKED ISSUE #314: Replace "Waveform could not be generated" message
In case of "one video, no audio", the “waveform could not be generated” message is still shown but not useful in this case, and subtracts from the intuition that the user can still cut up the video to their liking. -Sonia

Arnei:
In many cases, a waveform generation error will be due to the video files having no audio. A flatline represents this more accurately than some text. Furthermore, users really should not have to worry about waveform generation errors in the first place.

Resolves #314

### DIFF
--- a/src/main/Timeline.tsx
+++ b/src/main/Timeline.tsx
@@ -393,7 +393,8 @@ const Waveforms: React.FC<{}> = () => {
       );
     } else if (waveformWorkerError) {
       return (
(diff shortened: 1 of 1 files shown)


lkiesow:
If anyone is wondering how this looks like
Screenshot from 2021-10-11 18-40-35
:

Issue #307, opened 2021-05-20
Active state of scrubber and timeline hardly visible when it gets focus while tabbing through the controls


ziegenberg:
When going through the focusable elements of the page, every element is neatly highlighted. Only when selecting the scrubber or the timeline, it is hardly visible. 

See the recording attached. Starting at 0:13 the scrubber and the timeline are focused several times using the tab key: https://user-images.githubusercontent.com/856916/118942002-9b5dba80-b952-11eb-9e03-1f3bd9642646.mp4

ziegenberg:
Two more things:

- When cutting the timeline, the current timeline part, where the scrubber is located, is slightly darker. The colour difference should not be so subtle but more clearly visible.

- When cutting the timeline, the current timeline part, where the scrubber is located, is slightly darker. When now tabbing through the other timeline parts, it's almost impossible to see the current selection. Question is: should the scrubber follow the active timeline part while tabbing through?

Arnei:
> Question is: should the scrubber follow the active timeline part while tabbing through?

I'm not a keyboard user, but I doubt, at least with the way it currently works. You have to tab back through all the segments to get to the scrubber, so automatically moving it may have unintended side effects. But this discussion is probably better left for a different issue.

Arnei:
With the redesign of the editor, and modern browsers highlighting components by themselves when using tab navigation, I'd consider this issue resolved. If there are still issues with contrast or highlighting, do reopen (or open a new issue).

PR #442, opened 2021-09-21
Improved highlighting of segments

Arnei:
Replaces the outline for segments hovered by the scrubber. This should help with visual clarity (the dashed line was really not clear on that front).

Furthermore, adds an outline if the segment is in focus, as is current standard in Chrome and derivatives. This aims to help users relying on keyboard controls.

Helps with #307


Arnei:
Closing this because it is outdated

Issue #296, opened 2021-05-14
Workflow description does not parse as description


Arnei:
Why is the detail info for “publish” at the bottom of the list (makes it seem like a third option) and not next to the button? And why doesn’t “republish metadata” get detail info? - Sonia

Arnei:
"Republish Metadata" lacks a description because Opencast did not provide one. In such a case, maybe a text like "No description available" should be shown.

PR #444, opened 2021-09-21
Re-design of workflow selection

Arnei:
This patch overhauls the look and feel of the workflow selection in case users have to choose a workflow in the start processing step.

Bildschirmfoto vom 2022-04-22 11-12-23

Resolves #295
Resolves #296
Closes #445

PR #445, opened 2021-09-21
Display descriptive text instead of nothing for missing workflow description

Arnei:
Helps with #296


lkiesow:
Will be closed by #444

PR #516, opened 2021-12-02
Fix Build and Deploy Tests

lkiesow:
This patch fixes the `package-lock.json` to work with node version
v14.18.1 which is used in the GitHub Actions workflows.


lkiesow:
The partially removed web-vitals now also break the build.

lkiesow:
Already fixed as part of #518

PR #532, opened 2021-12-03
Changed font on metadata editor

### LINKED ISSUE #465: Different fonts used in metadata editor
All input fields should use the same font. They don't:

Screenshot from 2021-10-11 22-29-25

naweidner:
This fixes #465
Added textarea and input to index.css. Now the font and the size is the same. (It only affects metadata page)

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -1,4 +1,4 @@
-body {
+body, textarea, input {
   margin: 0;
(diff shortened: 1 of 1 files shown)


naweidner:
Without 'input' in css
without-input

With 'input' in css
with-input

lkiesow (approved):

PR #533, opened 2021-12-09
Centered Metadata-Page

### LINKED ISSUE #464: Homogenous layout
Most pages feature a centered layout. The metadata view is an exception. It is aligned left. It should also be centered.

naweidner:
This fixes #464
Centered formular, also added minWidth because the formular looked too small on full-width.


naweidner:
Formular on full-width
full-width

Formular on small-width
small-width

lkiesow (approved):
Seems to work as exptected

PR #547, opened 2021-12-14
Keyboard shortcut overview

### LINKED ISSUE #310: [Feature] Provide an overview of the available keyboard commands

naweidner:
This fixes #436  
This fixes #310 

1. "Play/Pause Video" now in one line
2. Button now sticky to bottom of navbar (navbar is now fixed in position)
3. Navbar is now covering whole page-height (sticky and scrollable, if height ist too small)
4. -
5. Layout is now centered and responsive, content is not overlapping anymore on small widths

Full-Width
full-width

Half-width
half-width

Small-width
small-width


lkiesow (changes requested):
Looks good. My only complaint is that we are missing a title tag. Can we add that for the keyboard shortcut button?

src/main/MainMenu.tsx
```diff
@@ -34,6 +34,9 @@ const MainMenu: React.FC<{}> = () => {
     flexShrink: 0,
     alignItems: 'center',
     padding: '20px',
+    height: '95vh',
+    position: 'fixed',
```
Arnei, line 38: Minor thing, but this causes the right borderline of the main menu to not line up precisely with the bottom of the page anymore.

Arnei:
> Here is a screenshot of my changes. The borderline of the main menu now lines up with the bottom of the page an we have a title tag.
> 
> keyboard-controls

I'm actually not a fan of text on this button, it makes it look to similar to the "real" main menu options. I think it should there should be clear visual distinction that the keyboard button is not directly related to the "main" buttons that pertain to the "main workflow" of the editor.

Ideally the control for accessing keyboard information would not be in the main menu side bar at all, but in a header or something like that. Alas, we don't have that, so this is the next best thing.

lkiesow:
@Arnei, what do you think about my suggestion from above? A title tag identical to the aria label?

src/main/MainMenu.tsx
lkiesow: For me, this seems to work better than before in both Firefox and Chromium.

Arnei:
> @Arnei, what do you think about my suggestion from above? A title tag identical to the aria label?

@lkiesow I ignored that suggestion initially because it really confused me and I only now have realized that you are talking about title attributes. I am absolutely in favor of adding that.

naweidner:
The button now looks like this. Removed the title and made the background (when this menu is active) a bit smaller

Shortcut-Button

PR #614, opened 2022-02-09
Basic view for selecting a subtitle to edit

### LINKED ISSUE #582: Create view to select subtitle track
- Configuration to map Names to flavors. Example:
    - `Deutsch` → `subtitle/editor+de`
    - `Englisch` → `subtitle/editor+en`
    - …
- User sees all existing subtitles and can select them for editing
- User can add new subtitles
    - Only one subtitle per language configuration
- Idea: Tile view with existing subtitles and a “plus” tile

Arnei:
- Resolves #582 

Adds a new main menu option "Subtitles" and a view where you can select a subtitle to edit or add based on a predefined list.


lkiesow:
A few initial thoughts:

- Not sure if it's just me, I think it's a bit confusing to use a Closed Captions icon but then use the term subtitles for the button description
- I probably do not care about the segment count on the overview page and would not put it there
- Maybe use some CSS like `text-overflow: ellipsis;  overflow: hidden;` for very long titles?

Arnei:
>     * Not sure if it's just me, I think it's a bit confusing to use a Closed Captions icon but then use the term subtitles for the button description

Maybe we should ask a couple more people opinions. Otherwise, this is the most accurate free fontawesome icon I found .D

>     * I probably do not care about the segment count on the overview page and would not put it there

The idea here was to give the user a sense of how finished each of the subtitles are, without them needing to click on each one to check. I don't mind dropping this either way, it was just an idea.

>     * Maybe use some CSS like `text-overflow: ellipsis;  overflow: hidden;` for very long titles?

Seems like the easiest solution, given that we don't expect many long titles anyway.

Arnei:
Closing in favor of https://github.com/opencast/opencast-editor/pull/638, where this PR is included in.

PR #619, opened 2022-02-24
Relax Playwright tests for the navigation bar

Arnei:
Currently the Playwright tests for the navigation bar expect four specific menu items to be present. However, every menu item besides "Cutting" and "Finish" is configurable to be present or not. Therefore I'd like to propose to only check for the actually required menu items, instead of some arbitrarily defined set.

### DIFF
--- a/tests/navigation.test.ts
+++ b/tests/navigation.test.ts
@@ -7,19 +7,12 @@ test('Test: Navigation', async ({ page, baseURL }) => {
     await expect(page).toHaveTitle("Opencast Editor");
 
     // checks if Navbar on left has 4 elements
-    const locator = page.locator('#root > div > div > nav > li');
-    await expect(locator).toHaveCount(4);
-
-    // Navigation to Metadata
-    await page.click('li[role="menuitem"]:has-text("Metadata")');
-
-    // Navigation to Tracks
-    await page.click('li[role="menuitem"]:has-text("Tracks")');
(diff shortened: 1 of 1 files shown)


tests/navigation.test.ts
```diff
@@ -7,19 +7,12 @@ test('Test: Navigation', async ({ page, baseURL }) => {
...
     // checks if Navbar on left has 4 elements
-    const locator = page.locator('#root > div > div > nav > li');
-    await expect(locator).toHaveCount(4);
-
-    // Navigation to Metadata
-    await page.click('li[role="menuitem"]:has-text("Metadata")');
-
-    // Navigation to Tracks
-    await page.click('li[role="menuitem"]:has-text("Tracks")');
+    const length = await page.locator('#root > div > div > nav > li').count();
```
lkiesow, line 10: Would it make things easier to add an HTML attribute like `data-testid="mainNavButton"` for selection instead of navigating through the whole DOM tree?

lkiesow (approved):
Looks good and should still break if there is no navigation any longer. I have one potential improvement below, but it's honestly good to go like this if you want it to.

tests/navigation.test.ts
Arnei: Probably would be more robust. On the other hand, I'm not sure about introducing ids just for testing. Is that common practice?

Arnei:
Thanks, I'm gonna merge this then and leave the optional improvement for another PR :)

PR #621, opened 2022-02-28
Made waveform always fill available height

Arnei:
The waveform appears flatter than it should be on lower resolutions. This fixes that by always having the waveform img consume the maximum available height.


lkiesow (approved):
Looks good. Can you fix the merge conflicts?

Screenshot from 2022-06-14 20-24-29

PR #638, opened 2022-03-02
SubtitleEditor

Arnei:
Includes all previous PRs for the subtitle editor.
<hr>

This PR aims at integrating the different views of the subtitle editor with the parsed subtitles from Opencast.
- The add subtitles button in the select view now has functionality
- The subtitle select view only shows buttons for languages that already exist in Opencast or have been newly created
- Displays error message for subtitles that could not be parsed (TODO: Make this more user friendly)


geichelberger:
Feedback after using the Subtitleeditor the first time:

- The curser looks different and cannot be moved, which is pretty annoying the first time.
grafik VS grafik
If you don't want the cursor to be set to a specific position, it could have the same shape without the part indicating you can grab it.
- Timeline: You cannot scroll horizontally (at least it did not work on firefox/fedora 36) 
- Timeline in focus: You cannot add a segment with keyboard shortcuts
- Input fields do not support dark mode
grafik
grafik

lkiesow:
I'll write it down here as well as on the backend (saving subtitles) pull request since it touches both parts:

I did only briefly skim through the code, but if I interpret the code correctly, subtitles are loaded from the internal publication, but then saved to the asset manager. This means that you don't necessary get what you save the next time you open the editor. It also means that you may overwrite completely different subtitle files and you might also miss subtitle files which are only in the asset manager.

My suggestion would be to always get the subtitles from the asset manager?

lkiesow:
Where are the flags coming from? Why is Generic Canadian? (…or Canada generic?)

Screenshot from 2022-08-15 14-10-30

lkiesow:
Another interesting question would be if we don't want to show all `captions/*` flavored tracks instead of hiding those not specifically configured. If they exist already, I would tend to show them. Hiding them could be confusing.

Arnei:
> Where are the flags coming from? Why is Generic Canadian? (…or Canada generic?)

The code for that is at `/src/main/SubtitleSelect.tsx` line 64-66 (can't figure out how to link to it). The flags are just kind of guessed at from the specified subflavor, as that usually includes the language code. Arguably that's a terrible approach, but I really thought the flags added to the visual.

> Another interesting question would be if we don't want to show all captions/* flavored tracks instead of hiding those not specifically configured. If they exist already, I would tend to show them. Hiding them could be confusing.

Iirc that behaviour was requested, though I do not remember the reasoning behind it.

ziegenberg:
Can we put this on the next Technical Meeting agenda?

naweidner:
I noticed some elements in darkmode are either very dark and hard to see or they don't support darkmode.

The big buttons are missing the `boxShadow` and `background-color`, since there is no more `finishMenuButtonStyle`. The new `tileButtonStyle` is missing these two attributes. This also affects the big buttons on the finish page. In darkmode the buttons are very hard to see.
grafik

Same here for the 'Pick a language' Button. 
Additionally the text on the dropdown-menus is hard to read in darkmode. 
grafik
The dropdown-menus are also missing darkmode colors.
grafik
grafik


Maybe give these buttons also a lighter `box-shadow` for darkmode? Also the inputfields are very bright in darkmode, maybe give them the same look, as the textbox and inputfields on metadata-page so the design is consistent?
grafik

naweidner:
What is the purpose of the Mini Timeline? 
At first I was confused why there is a white thick bar under the waveform (I am in darkmode) until I inspected it in the browser and saw it's a timeline.

What is this timeline for? Do we need it? And if so maybe you could make it look more like a timeline?

Arnei:
> Feedback after using the Subtitleeditor the first time:

Hey, thanks for the feedback and sorry for the late reply. Let me comment on some of the problems you faced

>     * The curser looks different and cannot be moved, which is pretty annoying the first time.

Agreed, having a similar look and feel to the cutting view only to work differently is confusing. Will go with your suggestion on that.

>     * Timeline: You cannot scroll horizontally (at least it did not work on firefox/fedora 36)

That is strange, it works for me on firefox/fedora 35. Do you perhaps have more details on that?
Ways to scroll horizontally:
1.  Clicking somewhere on the little grey bar at the very bottom of the page
2. Drag and dropping on the colored part of the timeline, below the subtitle segments (or any part of the timeline that is not a subtitle segment, really. Drag and dropping on a subtitle segment will try to move the subtitle segment instead).

>     * Timeline in focus: You cannot add a segment with keyboard shortcuts

That is currently arguably intended, as the keyboard shortcuts are specifically for "Adding a segment above/below (the current segment)". If you focus somewhere else, the editor would have to guess at where you want your segment to be added (at the end, at the beginning, at the segment that focused on last, relative to the segment currently hovered by the indicator on the timeline (which can technically be multiple segments)). Not sure what the intuitive behaviour would be in this case.

>     * Input fields do not support dark mode

PR so old it's older than dark mode. I'll try and get to that.

Arnei:
> What is the purpose of the Mini Timeline?

Since the "actual" timeline is strongly zoomed in, the mini timeline let's you get to other parts of the subtitle in a fast manner. I would argue this is useful, as the second fastest thing you could to achieve the same effect is scrolling the list view and then clicking on a cue. If you have a (naming) suggestion on how to make the mini timeline more intuitive, I'm all ears :D

ziegenberg:
Maybe call it "overview"? It gives you an overview of the whole timeline.

naweidner:
> the mini timeline let's you get to other parts of the subtitle in a fast manner

Ah yes that makes sense :D didn't thought of that.

> If you have a (naming) suggestion on how to make the mini timeline more intuitive, I'm all ears :D

'Mini Timeline' is good, maybe the left side of the marker could be colored? Like the timeline in Studio or the timeline on Youtube videos. 
grafik
grafik

ziegenberg:
Maybe this could be improved in the following way:

The mini timeline could be made a little bit higher (`20px` instead of `15px`) and the waveform could also be shown in the mini timeline. Additionally, you can show a lightly coloured "window" of the zoom displayed in the timeline above. Now, when you move the zoomed-in view it clearly shows where in the overall timeline you are.

I created a mockup.

Screen Shot 2022-09-14 at 13 33 34

geichelberger:
Instead of overlaying the subtitles, moving a subtitle over another one could change the duration accordingly. Likes it's solved in the subtitle editor of a giant video platform.

grafik
The first subtitle element gets "cut" after moving the blue one over it.
grafik

geichelberger:
This idea only makes sense if we do not want subtitle parts to overlay. I don't know if there is a use case for that.

Arnei:
> Instead of overlaying the subtitles, moving a subtitle over another one could change the duration accordingly. Likes it's solved in the subtitle editor of a giant video platform.

Agreed, that would be nicer. I also do not know if there is use case for overlapping segments. However, since the WebVTT spec allows it and the editor potentially has to deal with subtitle files from many different sources which may or may not use overlapping segments, I decided it was best to err on the side of caution and respect the spec, instead of forcing non-overlapping segments.

Arnei:
Regarding the suggestions for the mini timeline, I went with Nadines first idea of making it look more like a player bar for now. Mainly because the time I've been given to get this PR ready for approval is somewhat limited and Daniels approach would likely require some work to work well. For example, from the looks of it I would expect the highlight box to support drag and drop, but for long videos the area would get very small, so clicking should also still be supported. Also I'm assuming the center of the box denotes the current point in time? Then you might get some interesting behaviour when dragging the box to the very start or end of the video. In short, while it is not perfect, I hope the current MiniTimeline is good enough for approval ^^

ziegenberg:
I'll create a new Issue for my proposal for the Mini Timeline.

ziegenberg:
If we allow overlapping cue timings, it would be nice to have them as automatically stacking timeline elements, so it's clearly visible where the elements start and end.

Diggin around I also found resize handles to resize a single cue element on its left side. These resize handles should be more clearly visible. I propose to change them either to three vertical dots, as that's what I mostly see used when something is resizable. Or only show them as two chevrons, when hovering over the left edge of a cue element.

Both can be seen in the screenshot.

Screen Shot 2022-09-15 at 13 24 09

naweidner (approved):
Looks good. 
There is dark mode and the mini timeline looks better now.

geichelberger:
I think something went wrong with the last commit. https://test.editor.opencast.org/638/2022-09-16_12-42-28/ does not contain the subtitle tab.

geichelberger:
Oh, that was intentional, then maybe the config for the test deployments should be different?

ziegenberg:
Are you gonna squash those commits before merging? :)

Arnei:
> Are you gonna squash those commits before merging? :)

Not planning to. While it's a hell of a lot of commits, most of them actually do relevant stuff, and squashing them all together feels like grossly disregarding the development history.

lkiesow:
Sorry Arne, I found one more bug in the subtitle editor. But I hope this one is easy to fix:

Screenshot from 2022-10-11 16-56-03

HTML seems to be escaped, saved and then escaped when you load it again.

Arnei:
> HTML seems to be escaped, saved and then escaped when you load it again.

Do you have more details on that? I could not reproduce that.

Bildschirmfoto vom 2022-10-12 09-42-08

Although I could get the div tags to disappear after saving and reloading

Bildschirmfoto vom 2022-10-12 09-42-44

PR #650, opened 2022-03-22
Padding for long button labels

### LINKED ISSUE #649: Padding for long words
On the "Finish" page, the German translation of the buttons is very long. 
(especially "Änderungen speichern" and "Änderungen verwerfen")
Some padding should fix this.

Editor

naweidner:
This fixes #649
Added some padding left and right.

Editor-padding


naweidner:
> Could we instead pad without the width increase?

So do you mean by line break?

lkiesow:
Looks good:

``

PR #694, opened 2022-05-17
Generate waveform from track in specified flavor

Arnei:
Adds a new setting that allows the user to specify a flavor from which the waveform image will be generated. Since the waveform generation normally requires fetching a whole video file, this allows specifying a way smaller audio file, thereby reducing traffic.

<details>
  <summary>One way to generate an audio file in Opencast</summary>
  
  ## Example for an Opencast workflow operation for generating an audio file
    
    <operation
        id="encode"
        fail-on-error="true"
        exception-handler-workflow="partial-error"
        description="Encoding video">
      <configurations>
        <configuration key="source-flavor">presenter/preview</configuration>
        <configuration key="target-flavor">*/audio</configuration>
        <configuration key="target-tags">engage-download,engage-streaming,rss,atom</configuration>
        <configuration key="encoding-profile">audio.http</configuration>
      </configurations>
    </operation>

  ## Example for an Opencast encoding profile for a generating a small audio file
    profile.audio.http.name = convert video to audio
    profile.audio.http.input = visual
    profile.audio.http.output = audio
    profile.audio.http.suffix = .m4a
    profile.audio.http.ffmpeg.command = -i #{in.video.path} \
      -vn -filter:a aresample=8000 -b:a 12k -c:a aac -ac 1 \
      #{out.dir}/#{out.name}#{out.suffix}

</details>

Using a small audio file instead of a video file speeds up the waveform generation, but not as much as maybe desirable. For example, for a video file of 80 minutes and 125 MB, waveform generation took 20 seconds. With an audio file generated from the video with 7 MB, waveform generation still took 10 seconds.

This does NOT solve the issue of the editor crashing for very long videos (over ~3 hours). The crashing does seem to be related to workflow generation though, particularly [to audio decoding](https://github.com/opencast/opencast-editor/blob/3947a73656006c3ae6863625a3d08adc76746e70/src/util/waveform.js#L115).


lkiesow:
Do we still want/need this with the other one merged? We know that we still have problems with large files even with this merged. What do you think?

Arnei:
I think merging would still be beneficial for the current state of the editor.

True, this does not fix the issue with long files in frontend waveform generation. But currently, the only real solution to that seems to be to not use frontend waveform generation at all. By that logic, we should probably remove all code pertaining to waveform generation, but we haven't. Until we decide to do that, this PR helps users who want to use frontend waveform generation (and don't have long video files).

KatrinIhler:
Test failures. :(

gregorydlogan:
Per discussions at the technical meeting today, we're going to decline this for now.  If you want or need this please let us know and it can be refiled against a current branch.

PR #720, opened 2022-06-21
Added Darkmode

naweidner:
Added Darkmode for Editor.


ziegenberg:
In my opinion, there are two issues with accessibility on the metadata tab. It's the "Contributor(s)" and the "Created" Fields. See attached screenshots for details.

_It does not meet WCAG standards for accessible text, regarding colour and contrast. The current values are 3.59 and 3.79. For an AA rating, it needs to be at least 4.5 and for an AAA rating, it needs to be 7 or higher._

Screenshot from 2022-06-21 18-00-51
Screenshot from 2022-06-21 18-01-00

src/redux/themeSlice.tsx
```diff
@@ -0,0 +1,27 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+export const getTheme = () => {
+  const theme = localStorage.getItem('theme');
```
Arnei, line 4: Maybe call this `themeId` or something similar, to disambiguate it from the theme objects in cssStyles.tsx.

ziegenberg:
Thanks for the changed contrast values! This looks good now.

I only looked at the visual representation, not at the code.

Arnei (changes requested):
* Buttons seem to generally have become bigger. I recall that when first designing the editor, many were complaining that I was always making the buttons too big, so we probably want to avoid making them bigger again :D
* The DateTimePicker modals (for e.g. "Start Date") have no dark mode.
* There is an error in the console that looks new to me: `Using kebab-case for css properties in objects is not supported. Did you mean WebkitTextFillColor?`

lkiesow:
I don't like the really big mode toggle as a first and thus very highlighted option in the main menu. It works completely different than all other buttons by not switching any views, and you will barely ever need this. Maybe we could do something like replacing the keyboard icon with one for additional settings and views (represented e.g. by `···`) and then place the mode toggle and the keyboard settings there.

Also, +1 for smaller icons.

Arnei:
Correction: I can't seem to find why I thought the buttons where bigger, apparently they aren't. Forget that comment.

Arnei (commented):
* Very small thing in the DateTimePicker Modal: The forward/back arrows in the TimerPicker portion are not greyed out properly in dark mode. But I honestly wouldn't block this PR just because of that.
*Probably not your fault, but the icon on the settings button jumps slightly when hovered

Arnei (approved):

src/main/ThemeSwitcher.tsx
```diff
@@ -14,14 +14,15 @@ const ThemeSwitcher: React.FC<{}> = () => {
...
-    })
+    dispatch(toggleTheme(theme))
   }, [ themeState, theme, dispatch ] )
 
+  // Listen to system preference changes
+  const isDarkPrefered = window.matchMedia('(prefers-color-scheme: dark)');
+  isDarkPrefered.addEventListener('change', () => {
+    dispatch(toggleTheme(theme))
+  })
+
```
Arnei, line 25: Now the code is run on every rerender. Does this need to be run more than once on start-up?

src/redux/themeSlice.ts
```diff
@@ -66,14 +67,14 @@ export const themeSlice = createSlice({
     setState: (state, action: PayloadAction<theme["value"]>) => {
       state.value = action.payload;
     },
-    setTheme: (state, action: PayloadAction<theme["theme"]>) => {
+    toggleTheme: (state, action: PayloadAction<theme["theme"]>) => {
```
Arnei, line 70: Still taking a parameter it does not use

Arnei (changes requested):
Just happened to notice that the "Generating waveform" text and loading icon are still black in dark mode and thus very hard to see (they're in the middle of the waveform area).

src/main/ThemeSwitcher.tsx
```diff
@@ -1,71 +1,92 @@
...
-      transform: 'scale(1.1)',
-    },
-    '&:active': {
-      transform: 'scale(0.9)',
-    },
-  });
+    alignItems: 'center',
+  }
+
+  const selectStyle = {
```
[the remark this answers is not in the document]
Arnei: I'd assume you could define a general select style in `cssStyles.tsx`, like for example `basicButtonStyle`.

Arnei (approved):

PR #737, opened 2022-07-01
Updates react-redux to 8.0.2

Arnei:
Closes #1176

Frankly, I don't know why we now have to explicitly export the store.dispatch type to use AsyncThunks and didn't have to before, but if it pleases Typescript *shrug*


naweidner (approved):
Looks good and seems to work.

ziegenberg:
Hi @Arnei, the "Closes #1176" reference does not seem to be correct

PR #759, opened 2022-08-02
Added high contrast mode

naweidner:
- added high contrast mode dark and high contrast mode light
- used different colors for buttons and calendar (for better differentiation)
- contrast colors have a contrast ratio of at least 7:1


lkiesow:
First impression is good.
One area where I think the contrast is notably lower than everywhere else in high contrast mode though is the waveform. In particular in the dark high contrast theme.

Arnei:
I'm unsure if the applied changes actually increased the contrast in the waveform region. At the very least, the white scrubber (in dark high contrast mode) now has lower contrast, right?

Out of curiosity, what is the reasoning for having colored outlines instead of black/white ones? How do they relate to contrast? They give me the impression of using link colors for things that are not links.

Does this PR also increase the compatibility with other dark contrast modes (e.g. Windows dark contrast mode, Firefox standard high contrast colors)? If so, which ones did you test and would you share results? I realize that getting the editor to work with every dark contrast mode there is is a bit much, but users should at least be able to navigate to the dropdown to change to their preferred mode. Or can the editor now somehow detect if high contrast is preferred? (Akin to ligh/dark mode)

naweidner:
>At the very least, the white scrubber (in dark high contrast mode) now has lower contrast, right?

Oh, you are right. Didn't noticed that before when I changed the colors for the waveform. I will fix that.

>Out of curiosity, what is the reasoning for having colored outlines instead of black/white ones?

At first I decided to use only white/black outlines. But I thought the buttons and the metadata fields looked too similar, so I choosed a color, same for the calendar. Most elements in high contrast modes seem to have a bright color. So I used colors for better differentiation between buttons and other elements.

I will test other colors / color combinations. Maybe no outlines for the navigation buttons but the active one has a colored background?

>Or can the editor now somehow detect if high contrast is preferred?

Yes, if the user sets the color mode to "System design", the Editor detects the preferred system design (eg. dark, light, or high contrast dark/light).

Before I implemented the high contrast modes, the system contrast mode took away colors, shadows and borders (eg. the background color of the buttons in the navigation, when selected) and the waveform had no colors (eg. the waveform had a white or black background).

Arnei:
If I set my firefox to "high contrast mode" (Going into settings > general > colors > manage colors > set to always), the timeline bugs out:
Bildschirmfoto vom 2022-08-17 11-36-21
Is that something this PR should handle?

> I will test other colors / color combinations. Maybe no outlines for the navigation buttons but the active one has a colored background?
Did you get around to testing? What were your results?

naweidner:
> If I set my firefox to "high contrast mode" (Going into settings > general > colors > manage colors > set to always), the timeline bugs out. Is that something this PR should handle?

I think this could be fixed with firefox 's attribute 'moz-contrast'. I keep that in mind and add a condition to test for.

> Did you get around to testing? What were your results?

It is a bit difficult to choose the right colors with enough contrast.

The 'problem' is, that in high-contrast-dark, I have a dark background and a light scrubber, so if the waveform is too bright the contrast with the scrubber is too low. If the waveform is too dark the contrast with the background is too low. 

The best solution so far (imo) could be this combination:

grafik

naweidner:
Update: 
>I think this could be fixed with firefox 's attribute 'moz-contrast'. I keep that in mind and add a condition to test for.

`moz-contrast` doesn't seem to work. I will try something else.

naweidner:
It seems that the colors cannot be changed when high contrast mode is activated in the operating system. 

This is my workaround for that: the scrubber line now has a transparent outline that is only visible when high contrast mode is activated in the operating system, but otherwise does not affect the selectable color schemes.

Example:

High contrast (dark) mode activated via the settings-page:
image

High contrast mode activated via operating system (with light mode via settings-page):
image

High contrast mode activated via operating system (with dark mode via settings-page):
image

Arnei:
While this arguably still isn't perfect (the scrubber head still looks weird, and deleted segments aren't striped), I understand that this is not easily solveable and will largely depend on the high contrast colors set by the user anyway.

Therefore I'd argue we can merge this as is, especially since the explicit (not system design) high contrast modes work well and should not be blocked by this.

Arnei (approved):

PR #760, opened 2022-08-17
Thumbnail Editor

Arnei:
Adds a new view which allows users to create thumbnails for their events.

Allows users to:
- Generate a thumbnail from a video at a selected point in time.
- Upload a thumbnail.
- Set the same thumbnail for all videos.
- Discard any changes they made.

Has a "simple" and a "professional" mode.
- Simple: For when you only care about one primary thumbnail.
- Professional: When you want to be able to change all the thumbnails in the subflavor.

Requires Opencast backend PR to function: https://github.com/opencast/opencast/pull/4086

### DIFF
--- a/src/main/Video.tsx
+++ b/src/main/Video.tsx
@@ -35,7 +35,7 @@ import { selectTheme } from "../redux/themeSlice";
  * Container for the videos and their controls
  * TODO: Move fetching to a more central part of the app
  */
-const Video: React.FC<{}> = () => {
+export const Video: React.FC<{}> = () => {
 
   const { t } = useTranslation();
 
@@ -83,13 +83,13 @@ const Video: React.FC<{}> = () => {
   return (
     <div css={videoAreaStyle}>
       <VideoHeader />
-      <VideoPlayers />
+      <VideoPlayers refs={undefined}/>
       <VideoControls />
     </div>
   );
 };
 
-const VideoPlayers: React.FC<{}> = () => {
+export const VideoPlayers: React.FC<{refs: any, widthInPercent?: number}> = ({refs, widthInPercent=100}) => {
 
   const videoURLs = useSelector(selectVideoURL)
   const videoCount = useSelector(selectVideoCount)
@@ -99,14 +99,24 @@ const VideoPlayers: React.FC<{}> = () => {
     flexDirection: 'row' as const,
     justifyContent: 'center',
     alignItems: 'center',
-    width: '100%',
+    width: widthInPercent + '%',
   });
 
   // Initialize video players
   const videoPlayers: JSX.Element[] = [];
   for (let i = 0; i < videoCount; i++) {
-    // videoPlayers.push(<VideoPlayer key={i} url='https://media.geeksforgeeks.org/wp-content/uploads/20190616234019/Canvas.move_.mp4' />);
-    videoPlayers.push(<VideoPlayer key={i} dataKey={i} url={videoURLs[i]} isPrimary={i === 0}/>);
+    videoPlayers.push(
+      <VideoPlayer
+        key={i}
+        dataKey={i}
+        url={videoURLs[i]}
+        isPrimary={i === 0}
+        ref={(el) => {
+          if (refs === undefined) return
+          (refs.current[i] = el)
+        }}
+      />
+    );
   }
 
   return (
@@ -141,6 +151,7 @@ export const VideoPlayer = React.forwardRef(
   const ref = useRef<ReactPlayer>(null);
   const [ready, setReady] = useState(false);
   const [errorState, setError] = useState(false);
+  const [isAspectRatioUpdated, setIsAspectRatioUpdated] = useState(false);
 
   // Callback for when the video is playing
   const onProgressCallback = (state: { played: number, playedSeconds: number, loaded: number, loadedSeconds:  number }) => {
@@ -164,6 +175,7 @@ export const VideoPlayer = React.forwardRef(
         h = (ref.current.getInternalPlayer() as HTMLVideoElement).videoHeight
       }
       dispatch(setAspectRatio({dataKey, width: w, height: h}))
+      setIsAspectRatioUpdated(true)
     }
   }
 
@@ -171,8 +183,8 @@ export const VideoPlayer = React.forwardRef(
   const onReadyCallback = () => {
     setReady(true);
 
-    // Update the store with video dimensions for rendering purposes
-    updateAspectRatio();
+    // // Update the store with video dimensions for rendering purposes
+    // updateAspectRatio();
   }
 
   const onEndedCallback = () => {
@@ -194,6 +206,10 @@ export const VideoPlayer = React.forwardRef(
     if (clickTriggered && ref.current && ready) {
       ref.current.seekTo(currentlyAt, "seconds")
       dispatch(setClickTriggered(false))
+    }
+    if (!isAspectRatioUpdated && ref.current && ready) {
+      // Update the store with video dimensions for rendering purposes
+      updateAspectRatio();
     }
   })
 
--- a/src/main/Video.tsx
+++ b/src/main/Video.tsx
@@ -41,8 +41,6 @@ const Video: React.FC<{}> = () => {
 
   // Init redux variables
   const dispatch = useDispatch()
-  const videoURLs = useSelector(selectVideoURL)
-  const videoCount = useSelector(selectVideoCount)
   const videoURLStatus = useSelector((state: { videoState: { status: httpRequestState["status"] } }) => state.videoState.status);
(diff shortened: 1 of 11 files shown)


src/main/Video.tsx
```diff
@@ -196,6 +222,23 @@ const VideoPlayer: React.FC<{dataKey: number, url: string, isPrimary: boolean}>
...
+    // Returns the data url
+    captureVideo() {
+      const video = ref.current?.getInternalPlayer() as HTMLVideoElement
+      var canvas = document.createElement("canvas");
+      canvas.width = video.videoWidth;
+      canvas.height = video.videoHeight;
+      var canvasContext = canvas.getContext("2d");
+      if (canvasContext !== null) {
+        canvasContext.drawImage(video, 0, 0);
+        return canvas.toDataURL('image/png')
```
geichelberger, line 237: If I click on `Generate`, I get the exception below, and nothing happens. I am on Fedora 35 with Firefox Version 103.0.
```
Uncaught DOMException: The operation is insecure. main.30cb8141.js:2
    captureVideo Video.tsx:237
    generate Thumbnail.tsx:26
    onClick Thumbnail.tsx:256
    React 11
    bind_applyFunctionN self-hosted:1355
    Qt self-hosted:1318
    unstable_runWithPriority scheduler.production.min.js:18
    React 3
    bind_applyFunctionN self-hosted:1355
    $t self-hosted:1318
```

geichelberger (commented):
The view seems a little bit ~messy~ `unruhig`.

- The placeholder for the thumbnail could have a fixed ratio, like 16:9.
- Maybe deactivate the buttons that are not usable if a thumbnail is not defined

grafik

naweidner:
I agree with @geichelberger 
> The placeholder for the thumbnail could have a fixed ratio, like 16:9.

The box could have more width or at least some padding left and right. The text is right at the border and looks a bit squeezed. 
<br/>
**Three other things I noticed:**

**First one:** Maybe you could make the buttons "Generate" and "Use for other thumbnails" the same size? (See picture, I adjusted the size for these two buttons above). When the buttons have the same size, they maybe look more uniform/consistent?

grafik

**Second thing:** When I generate a thumbnail, the thumbnail and the buttons are very far apart. If you use a zooming-software, you (maybe) don't see that they belong together. (Something I learned about accessibility :D )

**And the third thing:** The box is right at the bottom of the page. I think some `margin` would fix this.

grafik

src/main/Video.tsx
Arnei: That should be fixed now

Arnei:
Thanks for your feedback. That'll teach me to only test things in Chrome and think "Should I test other browers? Naah, it'll be fine".

geichelberger:
Firefox seems to be picky :) 


### **Firefox**
grafik
### **Chrome**
grafik

geichelberger:
And it ignores the ratio if no thumbnail is set, which is working in Chrome.
grafik

naweidner:
True. I know what you mean. Most of the time I only test in Firefox (as for my review). In Chrome it doesn't take the whole width, when you generated a thumbnail. This time I tested in Firefox and Chrome.

The buttons look way better now and the view in Chrome fits very well

There are more things I noticed:

- There is no support for darkmode -> _Chrome and Firefox_
- The buttons do not have the same width and alignment after generating thumbnails (didn't noticed it before, because I only generated one thumbnail) -> _only in Firefox_
- I know german words can be very long, the text of the navigation-button hits the edges. The word "Miniaturbilder" maybe is the best solution so far. I've thought about alternative words, but couldn't find a shorter word that fits. "Vorschaubilder" is longer and also doesn't fit in. -> _Chrome and Firefox_
- The box for the thumbnail is still small/slim -> _only Firefox_

Btw: I test with Fedora 35 and Firefox 103.0 and Chrome 104.0.

This is how it looks with Firefox:

grafik

grafik

naweidner:
Oh no, you were faster than me. Let me check and correct my review.

Arnei:
> There is no support for darkmode -> Chrome and Firefox

Could you be more specific? The page looks pretty dark in dark mode to me.

naweidner:
I updated my last review. It looks way better now in Firefox.

naweidner:
I am sorry for spamming reviews.
In Chrome, the text still goes over the edge. Maybe 2px is not enough.

Chrome:
grafik

Firefox:
grafik


Here I tried it with 8px instead:

Chrome:
grafik

Firefox: 
grafik

geichelberger:
And there is still a difference between Firefox and Chrome, the ratio is not correct in firefox, and the `Use for other thumbnails` button has two rows instead of one.

Firefox:
grafik

Chrome:
grafik

ziegenberg:
All other buttons in the editor have an accompanying icon. "Generate", "Upload", etc., are missing those icons. At first, I did not recognize them as buttons and thought they were placeholders where the thumbnails would appear.

I also suggest removing the greyish background. Makes it look like it's patched in.

Maybe go with some like this for dark mode.
Screen Shot 2022-08-30 at 16 41 00

And this way for light mode.
Screen Shot 2022-08-30 at 16 48 03

ziegenberg:
I like the feature!

Arnei:
Did away with the boxed look.

I would also like to have icons for the buttons, but I could not come up with fitting icons for every button and did not want to cause more confusion by only having icons on like half the buttons.

ziegenberg:
Looks better now! Still, the buttons are bigger than any other button on the interface.

I propose to enlarge the thumbnail and show the buttons in a column along the thumbnail. To make better use of the whitespace around the two thumbnail/button boxes I suggest putting the assembly in its own flexbox and making it a row. With flexbox-wrapping enabled this should nicely work with smaller screens to and automagically reflow the layout. The buttons then all should have the same width.

With the following three settings on the thumbnail, I got them to resize smoothly, keeping their aspect-ration:
```css
aspect-ratio: 16 / 9;
max-width: 533px;
width: 100%;
height: 300px;
```

ziegenberg:
The button text "Use for other thumbnails" is much longer than the other three buttons. Which makes it look a bit weird and difficult to style uniformly. I suggest using a single word, something along the lines of:

- _Sync_
- _Synchronize_
- _Unify_

There are tooltips in place for all the buttons to help with understandig their purpose.

ziegenberg:
> I would also like to have icons for the buttons, but I could not come up with fitting icons for every button and did not want to cause more confusion by only having icons on like half the buttons.

How about the following icons?

| Button | FA-Name | icon |
| -------- | ------- | --------| 
| Generate | `fa-camera` | camera-solid |
| Generate (alternative) | `fa-brush` | brush-solid |
| Upload | `fa-upload` | upload-solid |
| Synchronize | `fa-copy` | copy-solid |
| Discard | `fa-trash` | trash-solid |
| Discard (alternative) | `fa-circle-xmark` | circle-xmark-solid |

Arnei:
I agree with your assessment that the thumbnail preview ought to be bigger and the buttons can be smaller, so I tried to implement that. Although I tried to refrain from using `aspect-ratio`, from my experience it behaves too differently in different browsers to be reliable.

Arnei:
I personally don't mind the length of "Use for other thumbnails". I always like to try and make the UI more dynamic before resorting to shorten the text, as it ends up helping out with translations (especially the germans somehow always end up with words that are way too long). I also find "Use for other thumbnails" easier to understand than shorter alternatives.

But I can also see the case for a shorter version. Ultimately I don't care too much about the name of the buttons, maybe everyone could vote for their favourite version on crowdin?

ziegenberg:
I very much like it now!
Though one more small thing. There is a size difference between the placeholder (`355x200`) and an actual thumbnail (`498x280`).


Screen Shot 2022-09-05 at 17 33 41

Screen Shot 2022-09-05 at 17 33 44

ziegenberg:
Another thing I found. The buttons increase their size when hovering over them. That's OK. They also increase their size when clicking. That's also OK. But they stay in the increased size, after clicking and when no longer hovering over them. I have to click somewhere else for them to be restored to their original size.

ziegenberg:
The accessibility check only reports one issue. For the upload buttons an invisible `<input />` is created. This should probably have `aria-hidden="true"` set to hide this non-interactive content from the accessibility API.

Screenshot from 2022-09-06 12-27-50

ziegenberg (approved):
Looks good to me!

naweidner (approved):
Looks good. Great work.

I am missing german translations, but this should not stop this PR from being merged.

geichelberger:
The approach of creating the thumbnail client side has a considerable disadvantage. The thumbnail quality depends on the video editor preview, which usually has a lower quality/resolution.

ziegenberg:
> The approach of creating the thumbnail client side has a considerable disadvantage. The thumbnail quality depends on the video editor preview, which usually has a lower quality/resolution. I think that's not a show-stopper for this PR, but there is room for improvement.
I created a new issue for this comment: #814

PR #768, opened 2022-08-30
make TrackSelection respect theme selection for menu headings

ziegenberg:
The CSS property `border-bottom` for the menu headings on the Track Selection was hard coded to `1px solid white`, which works great for the dark theme, but not so much for the light theme.

Before:
Screen Shot 2022-08-30 at 17 03 41

After:
Screen Shot 2022-08-30 at 17 03 57


ziegenberg:
You may wanna squash those two commits before merging... :)

Arnei (approved):
I was wondering if there had been a line there in the past or if I just imagined it, but it seems there was a line after all! Thanks for bringing it back.

PR #788, opened 2022-09-12
Improved tooltips

### LINKED ISSUE #786: Improve tooltip for current time and duration
When hovering over the current video time, it says something like "current time and duration", while when hovering over the duration it says something completely different. It would be better if it said "current time" when hovering over the current playtime and "duration" when hovering over the duration.

naweidner:
- added new and themed tooltips from material-ui
- added, changed and removed some tooltip texts in language files
- removed some unnecessary title-attributes
- removed `leftArrowButtonText` and `rightArrowButtonText` for Calendar (`next Month` and `previous Month`) because they could not be styled

This fixes #786


src/i18n/locales/de-DE.json
```diff
@@ -27,15 +27,15 @@
         "previewButton": "Vorschaumodus",
         "previewButton-tooltip": "Überspringt gelöschte Segmente bei der Wiedergabe des Videos. Derzeit {{status}}. Hotkey: {{hotkeyName}}",
         "previewButton-aria": "Vorschaumodus aktivieren oder deaktivieren. Hotkey: {{hotkeyName}}.",
-        "playButton-tooltip": "Play-Button",
-        "time-duration-tooltip": "Wiedergabezeit und -dauer",
+        "playButton-tooltip": "Video abspielen",
+        "current-time-tooltip": "Aktuelle Zeit",
```
Arnei, line 31: Arguably does not matter in this case, but I think it might be nicer to submit changes to existing translation keys through crowdin, just to keep to fair to the translators. (Unless of course it makes sense to include the change in the pull request because e.g. the function associated with the translated text changed).

src/i18n/locales/de-DE.json
```diff
@@ -44,8 +44,7 @@
     },
     "save": {
         "headline-text": "Aktuelles Projekt speichern",
-        "confirm-button": "Ja, Änderungen speichern",
-        "confirmButton-default-tooltip": "Speichern-Button",
+        "confirm-button": "Ja, meine Änderungen verwerfen",
```
Arnei, line 47: This translation is not correct and unnecessary?

src/main/Discard.tsx
```diff
@@ -65,15 +65,15 @@ const DiscardButton : React.FC<{}> = () => {
...
-    <div css={[basicButtonStyle, navigationButtonStyle(theme)]} title={t("discard.confirm-tooltip")}
-      role="button" tabIndex={0}
-      onClick={ discard }
-      onKeyDown={(event: React.KeyboardEvent<HTMLDivElement>) => { if (event.key === " " || event.key === "Enter") {
-        discard()
-      }}}>
-      <FontAwesomeIcon  icon={faTimesCircle} size="1x"/>
-      <span>{t("discard.confirm-button")}</span>
-    </div>
+      <div css={[basicButtonStyle, navigationButtonStyle(theme)]}
```
Arnei, line 68: Forgot the ThemedTooltip wrapper here I suppose?

src/main/Save.tsx
```diff
@@ -75,7 +76,7 @@ const Save : React.FC<{}> = () => {
   }
 
   return (
-    <div css={saveStyle} title={t("save.saveArea-tooltip")}>
+    <div css={saveStyle}>
```
Arnei, line 79: Might as well remove the translation key from the translation files.

src/main/Save.tsx
```diff
@@ -110,7 +111,7 @@ export const SaveButton: React.FC<{}> = () => {
   // Update based on current fetching status
   let icon = faSave
   let spin = false
-  let tooltip = t("save.confirmButton-default-tooltip")
+  let tooltip = ""
```
Arnei, line 114: Why set the tooltip to the empty string? What does this accomplish?

src/i18n/locales/de-DE.json
naweidner: I submitted the change to crowdin.

src/main/Discard.tsx
naweidner: There was one, but I removed it again and must have forgotten to undo the indentation.

src/main/Save.tsx
naweidner: Mainly two reasons:

- The first reason: the initial tooltip for the save-button would be "Save Button" and would be the only tooltip for this kind of buttons, this would not be consistent. When it`s empty, there is no tooltip shown. So it's partly a design decision to 'remove' this tooltip (also a tooltip should contain additional information to the user not repeating the elements text)
- The second reason for the empty string is, we need the variable and the title-attribute for the tooltip-wrapper has to be a string.

I will set the variable to null and then check for it in the tooltip-wrapper. If it's still null I will set it to an empty string

Arnei: > the initial tooltip for the save-button would be "Save Button" and would be the only tooltip for this kind of buttons, this would not be consistent. 

Makes sense, I can get behind that.

> The second reason for the empty string is, we need the variable and the title-attribute for the tooltip-wrapper has to be a string.

In all other cases, if a tooltip should not be displayed, the tooltip-wrapper is simply omitted (which is preferred I think, less visual clutter). Why do it differently in this case?

naweidner: > In all other cases, if a tooltip should not be displayed, the tooltip-wrapper is simply omitted (which is preferred I think, less visual clutter). Why do it differently in this case?

The button has multiple tooltips (for failing, loading and success), based on the workflowstatus. (I'm not sure why).
In my opinion we can also remove them. Then we don't need the tooltip-wrapper for this class.

Arnei: Aaah right, forgot about that. Maybe you could check if that is somehow relevant for screenreader folks. If not, all the tooltips can go. Makes no sense to remove one and keep the others.

src/i18n/locales/de-DE.json
ziegenberg: This translation is wrong and is updated via Crowdin.

src/i18n/locales/de-DE.json
```diff
@@ -27,15 +27,16 @@
         "previewButton": "Vorschaumodus",
         "previewButton-tooltip": "Überspringt gelöschte Segmente bei der Wiedergabe des Videos. Derzeit {{status}}. Hotkey: {{hotkeyName}}",
         "previewButton-aria": "Vorschaumodus aktivieren oder deaktivieren. Hotkey: {{hotkeyName}}.",
-        "playButton-tooltip": "Play-Button",
-        "time-duration-tooltip": "Wiedergabezeit und -dauer",
+        "playButton-tooltip": "Video abspielen",
+        "pauseButton-tooltip": "Video pausieren",
+        "current-time-tooltip": "Aktuelle Zeit",
+        "time-duration-tooltip": "Videodauer",
```
ziegenberg, line 33: Updates to translations should only happen via Crowdin, this prevents merge conflicts from happening. Only the source language (`en-US`) should be altered in Pull Requests.

ziegenberg (changes requested):

src/main/Metadata.tsx
```diff
@@ -544,6 +545,8 @@ const Metadata: React.FC<{}> = () => {
                 onBlur: (e: any) => {blurWithSubmit(e, input)},
                 showError: showErrorOnBlur
               }}
+              leftArrowButtonText={i18next.t('metadata.calendar-prev-month')}
+              rightArrowButtonText={i18next.t('metadata.calendar-next-month')}
```
Arnei, line 549: This will display the month text for all left/right arrows, including the ones to switch between hours and minutes

Arnei (changes requested):

Arnei (approved):

PR #792, opened 2022-09-15
Changed "saved successfully" to suggest that the editor can be closed now

Arnei:
It has been brought to my attention that, unlike me who often continues using the editor after clicking on "save", most normal users would want to exit the editor in some fashion after they have saved their work. This PR changes the post-save text to indicate that is now safe to close the editor. This also brings the text more in line with the post-process and post-discard text which both have this hint already.


owi92 (approved):
Looks good to me.

PR #794, opened 2022-09-28
Enable running the 'Update translations' workflow manually

ziegenberg:
This allows updating the translations manually just before cutting a release.

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -2,7 +2,8 @@ name: Update translations
 
 on:
   schedule:
-  - cron: "0 5 * * 1"
+    - cron: "0 5 * * 1"
+  workflow_dispatch:
 
 jobs:
   update-translations:


Arnei (approved):
Updating translations prior to cutting a release makes sense to me.

Should update the ReadMe in the future with this.

I wonder, how is this done in `opencast/opencast`? That has a fairly similar workflow, right?

ziegenberg:
I would also prefer to prevent contributors from editing translations outside of crowdin. This would cut down the number of possible merge conflicts on open PRs.

ziegenberg:
Sorry, I fat-fingered the wrong button.

ziegenberg:
`opencast/opencast` has a pretty similar workflow. See here for details: https://github.com/opencast/opencast/blob/develop/.github/workflows/update-translations.yml

Arnei:
> `opencast/opencast` has a pretty similar workflow. See here for details: https://github.com/opencast/opencast/blob/develop/.github/workflows/update-translations.yml

Right, so what I was wondering is, since that workflow also does not have `workflow_dispatch`, how do the Opencast release managers ensure an up-to-date translation when cutting releases. If it's better, we could adopt it, if it's worse, they could adopt this.

ziegenberg:
The Opencast [Release Manager Guide](https://docs.opencast.org/develop/developer/release-manager/#release-manager-guide) leaves you a bit in the dark about this. It's not overly specific...

gregorydlogan:
IIRC, Crowdin pushes changes nightly, or at least regularly.  The assumption on the RM side has been that the translations are current enough to cut a release at any point, and fixes would appear in later point releases anyway.

ziegenberg:
The crowdin updates are updated via a cron-ed weekly GHA workflow.

PR #795, opened 2022-09-28
Add workflow to check proper lable usage

ziegenberg:
The section "How to cut a release for Opencast" in the README.md states in the first section:

> Make sure all merged pull requests have proper type: labels. They are
> used for generating the release notes.

With this new github action this step gets automated.

### DIFF
--- a/.github/workflows/check-label.yml
+++ b/.github/workflows/check-label.yml
@@ -1,0 +1,23 @@
+# This workflow runs on any change made to a pull-request and aims to verify
+# that the correct label is present.
+
+name: Check proper lable usage
+
+on:
+  pull_request_target:
+    types: [opened, labeled, unlabeled, synchronize]
+
+jobs:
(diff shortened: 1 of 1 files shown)


Arnei (approved):
I like this and from what I could test it seems to work just fine, so approved!

With this action, we can probably remove the line you mentioned from the ReadMe?

ziegenberg:
Yes, we should be able to remove that line. And that was the purpose of this PR. ;)

Arnei:
> Yes, we should be able to remove that line. And that was the purpose of this PR. ;)

Okay, just asking because the PR does not actually remove the line :D

ziegenberg:
I would have then created a new PR once this change proved it really worked.

Arnei:
> I would have then created a new PR once this change proved it really worked.

... I guess that is fair enough for a gh-actions PR.

PR #827, opened 2022-10-13
Default subtitle flavor configuration

lkiesow:
This patch provides a default subtitle flavor configuration which should work with many installations out of the box.

More than that, it should never cause any trouble because the subtitle editor by itself is still hidden by default.

### DIFF
--- a/public/editor-settings.toml
+++ b/public/editor-settings.toml
@@ -113,8 +113,9 @@ password = "opencast"
 #
 # Example:
 [subtitles.languages]
-# "captions/source+de" = "Deutsch"
-# "captions/source+en" = "English"
+"captions/vtt+en" = "English"
+"captions/vtt+de" = "German"
+"captions/vtt+es" = "Spanish"
 
 # Specify the default video in the subtitle video player by flavor
 # If not specified, the editor will decide on a default by itself


Arnei (approved):

mtneug:
The `captions/source+<lang>` flavor idea comes from us. As these subtitles are still source files in the sense that they are tight to the source video and need to be trimmed together with the video, we applied the same idea of having a `captions/delivery+<lang>` flavor. What do you think of using that as the new default for Opencast?

lkiesow:
I think that's a great idea for the future. 
But this isn't changing the default, and the `captions/source+<lang>` that are removed are just comments.

I could already set that now, but at the moment, adopters are more likely to use `captions/vtt`.

If we want `captions/source+<lang>` to actually work (+1 for that), we also need to change more things in Opencast. In particular, the default workflow would actually need to cut subtitle files and we would need to publish only the ones we cut.

public/editor-settings.toml
```diff
@@ -113,8 +113,9 @@ password = "opencast"
 #
 # Example:
 [subtitles.languages]
-# "captions/source+de" = "Deutsch"
-# "captions/source+en" = "English"
+"captions/vtt+en" = "English"
+"captions/vtt+de" = "German"
+"captions/vtt+es" = "Spanish"
```
geichelberger, line 118: Shouldn't the default be the language code opencast uses in its metadata with three letters `eng, deu` instead of `en,de`

lkiesow: I don't see why. They are unrelated, aren't they?

lkiesow:
My original intention for this was mostly to provide a default configuration which works against the data on develop.opencast.org so that users can actually test the subtitle editor easily. But #828 provides a separate configuration for that anyway.

I see that just having a default set in here seems surprisingly controversial. That's why I'm declining this for now. But I'm going to start a broader conversation of how subtitles should work in Opencast and how we should apply flavors and tags. Once that discussion comes to a conclusion, we can revisit this.

public/editor-settings.toml
geichelberger: For the sake of consistency? If we choose a language classifier we should use it everywhere, else we would have to convert them all the time. I mean we can change the used classifier for the dublincore language term

e.g. The vosk workflow decides based on the Metadata which language to use and can add that as sub flavor extension.

PR #828, opened 2022-10-13
Improve GHA

ziegenberg:
This improves and fixes several GHA workflows.

### customized editor-settings.toml for deployments

Add a `editor-settings.toml` just for deployments to editor.opencast.org and test.editor.opencast.org. This enables setting certain config values for testing deployments we do not want to ship (yet).

### Update GitHub-actions

- Update to actions/checkout@v3 and actions/setup-node@v3
- Run GitHub Actions on the latest ubuntu

### fix GHA workflow 'cleanup deployment'

The steps were running in different working directories. Now they all run relative to GITHUB_WORKSPACE

### add check for opencast developers team membership for deployments

Currently, the Test & Deploy workflow only runs for selected pull requests. Adding a bunch of people to the list and keeping it up-to-date could turn out to be cumbersome. So this adds an additional workflow job and uses `TheModdingInquisition/actions-team-membership@v1.0` to check the membership of the person opening the PR. If the person is part of the `opencast/developers` team, the workflow will run.

**The `actions-team-membership` action needs a new Personal Access Token `GITHUB_ORG_TOKEN` with the `read:org` permission.**

### Check for updates to translations outside of Crowdin

This is done for two reasons:
1. This prevents possible merge conflicts when translations on Crowdin get updated.
2. As we are only uploading translation sources and not the translations themself to Crowdin, the translations on Crowdin and here in the repository would diverge and local changes would be overwritten with the next Crowdin update.

### DIFF
--- a/.github/workflows/cleanup-deployments.yml
+++ b/.github/workflows/cleanup-deployments.yml
@@ -23,11 +23,11 @@ jobs:
         path: editor-test
     
     - name: remove old builds
-      run: rm -rf editor-test/${{ github.event.number }}
+      run: rm -rf "$GITHUB_WORKSPACE/editor-test/${{ github.event.number }}"
 
     - name: regenerate index.html
       run: |
-        cd ~/editor-test
+        cd "$GITHUB_WORKSPACE/editor-test"
         echo '<html><body><ul>' > index.html
         find . -maxdepth 2 -name '*_*' -type d \
           | sort -r \
@@ -37,5 +37,6 @@ jobs:
     - name: Commit changes
       uses: stefanzweifel/git-auto-commit-action@v4
       with:
+        repository: $GITHUB_WORKSPACE/editor-test
         branch: gh-pages
         commit_message: Delete old deployments from PR ${{ github.event.number }}
--- a/.github/workflows/cleanup-deployments.yml
+++ b/.github/workflows/cleanup-deployments.yml
@@ -26,8 +26,8 @@ jobs:
       run: rm -rf "$GITHUB_WORKSPACE/editor-test/${{ github.event.number }}"
 
     - name: regenerate index.html
+      working-directory: $GITHUB_WORKSPACE/editor-test
       run: |
-        cd "$GITHUB_WORKSPACE/editor-test"
         echo '<html><body><ul>' > index.html
         find . -maxdepth 2 -name '*_*' -type d \
           | sort -r \
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -11,10 +11,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: use node.js 16.x
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: 16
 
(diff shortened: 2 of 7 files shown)


.github/workflows/cleanup-deployments.yml
```diff
@@ -23,11 +23,11 @@ jobs:
         path: editor-test
     
     - name: remove old builds
-      run: rm -rf editor-test/${{ github.event.number }}
+      run: rm -rf "$GITHUB_WORKSPACE/editor-test/${{ github.event.number }}"
 
     - name: regenerate index.html
       run: |
-        cd ~/editor-test
+        cd "$GITHUB_WORKSPACE/editor-test"
```
lkiesow, line 30: We could also use

```yaml
working-directory: $GITHUB_WORKSPACE/editor-test
```

ziegenberg: Suggestion adopted

ziegenberg:
Many of the dev deployments for the subtitle editor had the subtitle editor turned off because that was what the default was set to. And that's still the case.

To my understating the `public/editor-settings.toml` file is the one config file meant to be released. And it also kinda serves as documentation of the available config options. And this file currently has the following settings: 

```
# Username, used for HTTP basic authentication against Opencast.
# Not defining this will work just fine if integrated in Opencast.
# Type: string | undefined
# Default: undefined
name = "admin"

# Username, used for HTTP basic authentication against Opencast.
# Not defining this will work just fine if integrated in Opencast.
# Type: string | undefined
# Default: undefined
password = "opencast"
```

So `editor-settings.toml.release` and `editor-settings.toml.dev` have the only purpose to have a different config file for our editor.opencast.org showcase with everything enabled and to have a separate config file for test deployments.

ziegenberg:
You're suggesting the opposite of what #837 is doing?

lkiesow:
Basically, yes.
Right now, #837 makes sense since we only have one configuration and that also serves as documentation.
If we split the files apart and use `public/editor-settings.toml` for development anyway, I think it's good to use that for all the development relates stuff. Otherwise, we still have this configuration file service two purposes. We just add two more configuration files to maintain :D

ziegenberg:
Adding a `editor-settings.toml` to `/` with our current production recommendations and also publishing that alongside releases sounds reasonable.

Would you adapt your #837 to do that? I can change my PR afterwards.

ziegenberg:
@lkiesow I dropped the two commits adding the additional config files.

lkiesow:
Sorry for the delay. I'll also update the other pull request with the configurations later today… hopefully.

> The actions-team-membership action needs a new Personal Access Token GITHUB_ORG_TOKEN with the read:org permission.

How would you suggest this should work? Merging this right now should break the CI since no such token exists. Adding one requires a personal token. Which one would we use? The token we use would afaik provide access to all the users organizations. Also, since it's a personal token, that user could always “clean up” and thus accidentally break the CI couldn't he?

lkiesow:
@ziegenberg, I've created a pull request splitting up the configuration at #844

ziegenberg:
Hi @lkiesow!

> Sorry for the delay. I'll also update the other pull request with the configurations later today… hopefully.
> 
> > The actions-team-membership action needs a new Personal Access Token GITHUB_ORG_TOKEN with the read:org permission.
> 
> How would you suggest this should work? Merging this right now should break the CI since no such token exists. Adding one requires a personal token. Which one would we use? The token we use would afaik provide access to all the users' organizations. Also, since it's a personal token, that user could always “clean up” and thus accidentally break the CI, couldn't he?

I amended my PR and changed it to use the already available `secrets.GITHUB_TOKEN`.

PR #836, opened 2022-10-19
Tooltips for subtitle editor

### LINKED ISSUE #835: Tooltip classes are missing from some elements with title tags
…at least in the subtitle section:

index

Tooltips need to be applied similar to this:
- https://github.com/opencast/opencast-editor/pull/788/files#diff-8ab5f61171f48f99afe49140fb9874965930eb5ac0ac01b9014f669dd4874b32

naweidner:
- added tooltips for subtitle editor
- this fixes #835


src/i18n/locales/en-US.json
```diff
@@ -257,6 +257,10 @@
       "selectVideoLabel": "Video Flavors"
     },
 
+    "subtitleTimeline": {
+      "timeline-tooltip": "Mini Timeline"
```
lkiesow, line 261: - I don't think that `Mini Timeline` is a very good description of this interface element. The tooltip should explain what an element does or what it is for. Maybe use something like “Overview Timeline”?
- Given that the subtitle editor has two timelines, the identifier `subtitleTimeline.timeline-tooltip` doesn't explain which timeline this translation string is for.  Plus, it's mixing camel case and kebab case. Maybe use `subtitleTimeline.overviewTimelineTooltip`?

lkiesow (changes requested):
https://user-images.githubusercontent.com/1008395/198658012-2adb09f9-9c99-48e7-bee8-2d5cbc2d8e17.mp4

src/main/SubtitleListEditor.tsx
```diff
@@ -402,54 +396,90 @@ const SubtitleListSegment = React.memo((props: subtitleListSegmentProps) => {
...
           />
         </div>
-
         <div css={functionButtonAreaStyle} className="functionButtonAreaStyle">
-          <div css={[basicButtonStyle(theme), addSegmentButtonStyle]}
-            role="button" tabIndex={0}
-            title={t("subtitleList.addSegmentAbove")}
-            arial-label={t("subtitleList.addSegmentAbove")}
+          <FunctionButton
+            style={[basicButtonStyle(theme)]}
```
Arnei, line 401: Seems weird to have component to which you can pass a style, which is then partially overwritten in the component. I am curious what your rationale for that choice was? Do you expect different `basicButtonStyles` in the future?

naweidner: I'm no longer quite sure what I was thinking and why I did it the way I did it. But where you mention it, you're right, it doesn't make any sense. I will remove it and add it directly to the component, without the 'basicButtonStyle' e.g. the icons in the buttons are no longer centered and there is no hover transformation anymore.

Arnei (approved):

PR #838, opened 2022-10-19
Remove superfluos "=" on landing page

Arnei:
If you blindly copy and paste the query parameter into your browsers address window (and change the mp id), you just land on the landing page again, leaving you a bit stumped before figuring out the typo.


lkiesow (approved):

lkiesow:
This should probably have gone through the technical meeting since it wasn't urgent to get this in.

Arnei:
Very sorry. Since it was approved so fast I felt vindicated in my assumption that this PR would be absolutely non-controversial.

PR #840, opened 2022-10-20
Update language strings

geichelberger:
Updates thumbnail-related language strings and use positive wording when discarding changes.

The significant changes are replacing videos with tracks because the term is already introduced in the tab before and updating the explanation to be more encouraging.


Arnei (approved):

src/i18n/locales/en-US.json
```diff
@@ -168,24 +168,24 @@
...
       "title": "Thumbnail Editor",
-      "noThumbnailAvailable": "No Thumbnail set",
-      "previewImageAlt": "Thumbnail for",
+      "noThumbnailAvailable": "No Thumbnail available",
+      "previewImageAlt": "Thumbnail for track",
       "buttonGenerate": "Generate",
-      "buttonGenerate-tooltip": "Create a new thumbnail from the current scrubber position",
-      "buttonGenerate-tooltip-aria": "Create a new thumbnail from the current scrubber position",
+      "buttonGenerate-tooltip": "Generate a new thumbnail from the current scrubber position",
+      "buttonGenerate-tooltip-aria": "Generate a new thumbnail from the current scrubber position",
```
lkiesow, line 175: Maybe player is better since that's also where user will see the current image?

```suggestion
      "buttonGenerate-tooltip": "Generate a new thumbnail from the current player position",
      "buttonGenerate-tooltip-aria": "Generate a new thumbnail from the current player position",
```

geichelberger: I'm not too fond of the wording scrubber position in general. For me, that's s a timeline marker (position). But I am Adobe-biased :D

The player position sounds like the player's actual position on the screen, not the playhead position or current playback position.

cgruber-uos: For me "scrubber" is also a kind of "technical term" that some non nativ speaker has to learn. "timeline marker" sound for me more intuitive and could be easily deduced from "time" and "line". And I would expect that "timeline" is more used as "scrubber" in general. So more understandable for non natives. 

In case we have not use "scrubber" all the time on other places. I would count for  "timeline marker" also. And I'm no Adobe fanboy! :-)

geichelberger: Renaming all the scrubber occurrences is out of that PR's scope. I will file a new one.

PR #841, opened 2022-10-21
High contrast mode for subtitle editor

naweidner:
- subtitle view now has a working  high contrast mode
- added some new color-attributes (subtitle_segment_bg, subtitle_segment_border and subtitle_segment_text)
- colors for subtitle elements adjusted (bg-color, text-color, border-color etc)


lkiesow (changes requested):
The high contrast mode lowers the contrast for the subtitle editor's video selector. The label is not legible at all anymore and the border is also no longer visible, making it hard to understand that this is a dropdown menu:

Screenshot from 2022-10-28 17-02-13

In high contrast mode, the language selector is literally black on black and you have to mark the text to read it:

Screenshot from 2022-10-28 16-59-32


The borders of the segment start and end times run into each other in high contrast mode. There should be at least a bit of margin between these fields:

Screenshot from 2022-10-28 17-00-45

src/main/SubtitleVideoArea.tsx
```diff
@@ -185,11 +186,29 @@ const VideoSelectDropdown : React.FC<{
 
   const onSubmit = () => {}
 
-  const muiTheme = createTheme({
-    palette: {
-      mode: isDarkPreferred === 'dark' ? 'dark' : 'light',
+  const muiStyle = makeStyles({
```
Arnei, line 189: This is already defined in SubtitleSelect.tsx. Can we define it in cssStyles.tsx instead to avoid duplication?

Arnei (changes requested):
The issues above all seem to have been addressed. Got one more related to the code.

Arnei:
The hint for the dropdowns does not show if the dropdown was in focus once and then isn't anymore. Only in dark-high-contrast mode. Probably because the color matches the background-color.
[Screencast_30.11.2022_16:35:26.webm](https://user-images.githubusercontent.com/14070005/204840970-70ff66b2-4f83-422c-b83a-e93913eafcfa.webm)
(Same for the create dropdown in the select view)

naweidner:
Last commit didn't fixed the label-issue. Can't reproduce this on localhost (on localhost it looks "normal" for all color modes). 
This is only visible when you use a deploy link.

I also noticed that the clock in the calendar (in the metadata view) looks normal via the deploy link (all color modes) but the colors are not correct via localhost (all color modes).

localhost:
grafik

Deploy-Link - https://test.editor.opencast.org/841/2022-12-01_13-50-28/ :
grafik

Arnei:
Seems like running the app with `start` or building with `build` results in slightly different css-classes and, more importantly for us, different css-class-*names*. Since we hard-code css-class-names for the MUI elements, using a class name that is only present in one of the builds will result in the rule missing from the other.

It would still be extremely great if we didn't have to rely on hard-coding values for automatically generated class names.

Arnei:
Somewhat nitpicky, but in regarding the clock view in the calendar: In light high contrast mode, for the arrows to switch between hours and minutes, the contrast is not very high (they are black when you can't click them, and dark blue when you can).

Looks good otherwise. Very happy you managed to get rid of the hard-coded class names.

Arnei (approved):

Issue #813, opened 2022-10-05
Text for multi-select in Metadata view in dark mode hardly visible


Arnei:
In the metadata view, there are fields where you can select multiple items from a dropdown, and even add new ones by typing. When typing however, the text is hardly visible as it's grey and not white like the rest of the text.

PR #842, opened 2022-10-22
change text color for react-select multi CreatableSelect

ziegenberg:
The presenter and contributor fields in the metadata tab use a combined select input control with multi-select enabled. When typing the text is hardly visible in the dark theme, as it's grey on a dark background.

This commit sets the colour on the `input` component to `theme.text` to respect the theme selection.

For the list of available components for React-Select see: https://react-select.com/components

Fixes: #813

### DIFF
--- a/src/cssStyles.tsx
+++ b/src/cssStyles.tsx
@@ -262,5 +262,9 @@ export function selectFieldStyle(theme: Theme) {
       ...provided,
       cursor: "text",
     }),
+    input: (provided: any) => ({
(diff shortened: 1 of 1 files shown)


lkiesow:
This patch doesn't seem to make the issue any better:

Screenshot from 2022-10-28 15-58-49

ziegenberg:
Fixed it! Now it is way better.
Screenshot from 2022-10-28 18-44-28

PR #837, opened 2022-10-19
Set Defaults in Configuration

lkiesow:
This patch updates the main configuration file, adding the correct default values to the configuration options.

PR #844, opened 2022-10-25
Split demo and production configuration

lkiesow:
This patch splits the configuration file to provide two separate versions:

- `editor-settings.toml` which is our recommended configuration for production, has all the default settings and documenta all configuration options.

- `public/editor-settings.toml` which is used for development (`npm run start` will use this) and for demo deployments. It should enable new features by default to make testing easier.

This closes #837


ziegenberg (approved):
LGTM!

PR #845, opened 2022-10-25
fix Test & Deploy workflow

ziegenberg:


lkiesow:
Seems reasonable. Let's just see if this works…

ziegenberg:
oh, that was quick... :D

PR #848, opened 2022-10-25
fix Test & Deploy workflow

ziegenberg:
Secret names must not start with `GITHUB_ `.

### DIFF
--- a/.github/workflows/update-test.yml
+++ b/.github/workflows/update-test.yml
@@ -112,7 +112,7 @@ jobs:
         uses: TheModdingInquisition/actions-team-membership@v1.0
         with:
           team: developers # required. The team to check for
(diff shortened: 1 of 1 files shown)


.github/workflows/update-test.yml
```diff
@@ -111,12 +111,54 @@ jobs:
...
+                    nodes {
+                      name
+                  }
+                  pageInfo {
+                    hasNextPage
+                    endCursor
+                  }        
+                }
+              }
+            }
```
geichelberger, line 146: Alternative Query:
```
query($organization: String!, $username: String!, $team: String!){
  user(login: $username) {
    organization(login: $organization) {
      team(slug: $team) {
        id
        name
      }
    }
  }
}
```

PR #855, opened 2022-10-27
Add aria-label to theme selector

geichelberger:
The appearance selector, which is used to select themes for accessibility is missing an aria-label.


ziegenberg:
@geichelberger could you rebase your branch, so the deployment check can run?

Arnei (approved):

PR #860, opened 2022-10-28
Icon Configuration for Languages

### LINKED ISSUE #825: Inconsistent language flags
Given the subtitle configuration
```toml
[subtitles.languages]
"captions/vtt+de" = "Deutsch"
"captions/vtt+en" = "English"
```
you end up with a view like this:

Screenshot from 2022-10-13 16-14-23

- Where does the German flag come from?
- Why is there none for English?
- Can we make this somewhat consistent?

lkiesow:
This patch allows users to specify icons for languages selectors displayed in the subtitle editor. Users have to specify the selectable languages anyway and the algorithm used before guessed wrong for popular languages like English.

If no icon is specified for a language, the selector will simply omit the icon and display just the label.

This fixes #825

Screenshot from 2022-10-28 18-35-00

### DIFF
--- a/editor-settings.toml
+++ b/editor-settings.toml
@@ -81,6 +81,14 @@
 "captions/source+en" = "English"
 "captions/source+es" = "Spanish"
 
+[subtitles.icons]
+# A list of icons to be displayed for languages defined above.
+# Values are strings but should preferably be Unicode icons.
+# These are optional and you can also choose to have no icons.
+"captions/source+de" = "🇩🇪"
+"captions/source+en" = "🇺🇸"
+"captions/source+es" = "🇪🇸"
+
 [subtitles.defaultVideoFlavor]
 # Specify the default video in the subtitle video player by flavor
 # If not specified, the editor will decide on a default by itself
--- a/public/editor-settings.toml
+++ b/public/editor-settings.toml
@@ -32,6 +32,12 @@ mainFlavor = "captions"
 "captions/source+de" = "Deutsch"
 "captions/source+en" = "English"
 "captions/source+es" = "Spanish"
+"captions/source" = "Generic"
+
+[subtitles.icons]
+"captions/source+de" = "🇩🇪"
+"captions/source+en" = "🇺🇸"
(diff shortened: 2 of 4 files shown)


editor-settings.toml
```diff
@@ -81,6 +81,13 @@
 "captions/source+en" = "English"
 "captions/source+es" = "Spanish"
 
+[subtitles.icons]
+# A list of icons to be displayed for languages defined above.
```
Arnei, line 85: We could mention that these are totally optional? Arguably that might be obvious, but it might also not be. And in that case, no one feels pressured into thinking up icons when they don't want.

Alternatively comment the list out?

Arnei (approved):
I agree with this PR. Correctly guessing all the language flags was a pipe dream anyway.

This also allows us to avoid the discussion whether country flags are sensible indicators for languages, which is nice.

Comments below are justodds and ends.

Arnei:
One more comment: Since the "icons" can now be very long, it might make sense to update the icon css a bit, although I hope we can trust people who are allowed to set the config to choose sensible values :)
Bildschirmfoto vom 2022-11-11 15-05-07

editor-settings.toml
lkiesow: I'll update the description

lkiesow:
>  it might make sense to update the icon css a bit, 

I kind of feel like saying that if people want to be stupid, let them be stupid ;-)
But you were thinking about something like `overflow: hidden`?

Arnei:
> > it might make sense to update the icon css a bit,
> 
> I kind of feel like saying that if people want to be stupid, let them be stupid ;-) But you were thinking about something like `overflow: hidden`?

Yes, thinking about exactly that. Wouldn't spent any major on time this though if `overflow: hidden` for some reason does not work as expected or is not enough. Not like admins will intentionally configure the editor to look terrible, right?

lkiesow:
Updated the style. If you now try adding hundreds of flags, you will end up with something like this:

Screenshot from 2022-11-23 18-02-25

PR #882, opened 2022-11-02
Make active workflow warning less technical

lkiesow:
This patch updates the warning users get when an active workflow is running on the event that is being loaded in the editor.

- Technically, the error message was not correct since it's not the problem that a workflow is being processed in Opencast, but that this workflow is running on the event the user tries to edit.
- The wording was very technical. A non-technical user does not know what a workflow in the context of Opencast is and what is happening.

This patch changes the message to tell users that the event they want to edit is being processed. That should hopefully be easier to understand.


geichelberger (approved):
LGTM

Arnei (approved):

PR #884, opened 2022-11-02
Thumbnail for video streams only

### LINKED ISSUE #853: Thumbnail editor shows UI for non-video tracks
Thumbnail Editor may show UI for tracks that don't have a video or audio stream, for example subtitle tracks. It should not show UI for tracks that cannot reasonably have a thumbnail (Fairly sure there was a check for this during development at some point, but it seems it got lost).
Bildschirmfoto vom 2022-10-26 16-10-00

To reproduce, open the event with id "ID-subtitle-demo" from develop.opencast.org.

lkiesow:
This patch limits the thumbnail generator to only show a selection for video tracks, omitting things like audio or subtitle tracks.

The downside of this is obviously that pure audio streams aren't supported. But they aren't really supported in the current user interface anyway.

This fixes #853

### DIFF
--- a/src/main/Thumbnail.tsx
+++ b/src/main/Thumbnail.tsx
@@ -9,7 +9,7 @@ import { useDispatch, useSelector } from "react-redux";
 import { settings } from "../config";
 import { basicButtonStyle, deactivatedButtonStyle, flexGapReplacementStyle, titleStyle, titleStyleBold } from "../cssStyles";
 import { selectTheme, Theme } from "../redux/themeSlice";
-import { selectOriginalThumbnails, selectTracks, setHasChanges, setThumbnail, setThumbnails } from "../redux/videoSlice";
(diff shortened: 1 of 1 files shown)


src/main/Thumbnail.tsx
```diff
@@ -130,6 +130,7 @@ const ThumbnailTable : React.FC<{
 }> = ({inputRefs, generate, upload, uploadCallback, discard}) => {
 
   const tracks = useSelector(selectTracks)
```
Arnei, line 132:
```suggestion
  const tracks = useSelector(selectVideos)
```

How about using the `selectVideos` selector instead (and maybe rename `tracks` to `videos`)? Should effectively do the same thing.

lkiesow:
Switched to `selectVideos` and renamed `tracks` to `videoTracks`

Arnei (approved):

PR #885, opened 2022-11-02
Show all existing subtitles

### LINKED ISSUE #826: Show all existing subtitles
If your media pacage has a subtitle which is found by the backend and delivered to the frontend, the frontend will still not show it unless you have specifically configured the flavor in `subtitles.languages`. This can easily cause existing subtitles to get lost.

It would be great if _all_ existing subtitles could be shown in the subtitle editor view. Though, I think, it's fine that only configured languafes are shown in the drop-down for creating new subtitle tracks.

lkiesow:
This patch shows all existing subtitles that are recognized by Opencast and sent to the editor in the user interface instead of just showing subtitles which exactly match a specified language flavor.

Which subtitle tracks are sent from Opencast to the front-end is still configurable in the editor service.

This also does not allow users to create arbitrary subtitle tracks. They have to already exist.

This fixes #826

### DIFF
--- a/src/i18n/locales/en-US.json
+++ b/src/i18n/locales/en-US.json
@@ -238,7 +238,8 @@
       "backButton": "Back",
       "backButton-tooltip": "Return to subtitle selection",
       "editTitle": "Subtitle Editor - {{title}}",
-      "editTitle-loading": "Loading"
+      "editTitle-loading": "Loading",
+      "generic": "Generic"
     },
 
     "subtitleList": {
--- a/src/main/SubtitleSelect.tsx
+++ b/src/main/SubtitleSelect.tsx
@@ -18,8 +18,9 @@ import { createTheme, ThemeProvider } from '@mui/material/styles';
 /**
  * Displays buttons that allow the user to select the flavor/language they want to edit
  */
- const SubtitleSelect : React.FC<{}> = () => {
+const SubtitleSelect : React.FC<{}> = () => {
 
+  const { t } = useTranslation();
   const captionTracks = useSelector(selectCaptions) // track objects received from Opencast
   const subtitles = useSelector(selectSubtitles)    // parsed subtitles stored in redux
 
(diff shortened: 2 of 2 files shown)


src/main/SubtitleSelect.tsx
```diff
@@ -28,36 +28,26 @@ import { createTheme, ThemeProvider } from '@mui/material/styles';
...
+        subFlavor: flavor,
+        title: languages[flavor] || lang});
+      delete languages[flavor];
     }
 
+    // List of unused languages
+    let tempCanBeAddedFlavors = Object.keys(languages)
+      .map(flavor => ({subFlavor: flavor, title: languages[flavor]}));
+
     setDisplayFlavors(tempDisplayFlavors)
```
Arnei, line 51: Maybe we want to sort the flavors before setting them? Before, they where implicitly ordered by their order in the config file, due to iterating over `settings.subtitles.languages`. Now that implicit ordering is gone, it seems the ordering is less consistent. For example when viewing the subtitle select UI, if I have a subtitle track called "English" and add "Deutsch", "Deutsch" will be added to the right of "English". If I click on "Deutsch" and go back to the select view, it will now be to the left of "English".

Arnei (changes requested):

src/main/SubtitleSelect.tsx
```diff
@@ -28,36 +28,26 @@ import { createTheme, ThemeProvider } from '@mui/material/styles';
...
-        }
-      }
-
-      if (found) {
-        tempDisplayFlavors.push({subFlavor: subFlavor, title: name})
-      } else {
-        tempCanBeAddedFlavors.push({subFlavor: subFlavor, title: name})
-      }
+    for (const flavor of subtitleFlavors) {
+      const lang = flavor.replace(/^.*\+/, '') || 'Generic';
```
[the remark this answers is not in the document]
lkiesow: Do you think it's better to translate this or to make this configurable?

src/main/SubtitleSelect.tsx
lkiesow: Sort by `title`?

Arnei: Yes, that seems reasonable.

src/main/SubtitleSelect.tsx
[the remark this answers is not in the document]
Arnei: The only case that I can think of where a user would want to configure this is if they think 'Generic' does not properly convey that a title is missing. In which case the user should tell us so we can update the translation for everyone's benefit. So I think translation is fine ^^

lkiesow:
- Subtitle select buttons are now sorted by title
- The string `Generic` can now be translated

src/i18n/locales/en-US.json
```diff
@@ -238,7 +238,8 @@
       "backButton": "Back",
       "backButton-tooltip": "Return to subtitle selection",
       "editTitle": "Subtitle Editor - {{title}}",
-      "editTitle-loading": "Loading"
+      "editTitle-loading": "Loading",
+		"generic": "Generic"
```
Arnei, line 242:
```suggestion
	"generic": "Generic"
```

Arnei (changes requested):
Basically approved, seems there's just one small indentation issue.

src/i18n/locales/en-US.json
lkiesow: That shouldn't have happened. Sorry. Fixed.

Arnei (approved):

PR #894, opened 2022-11-14
Fix preview button tooltip hotkey hardcoded

Arnei:
Hotkey for the preview button was hardcoded and would thus show incorrectly for i.e. MacOS users.


lkiesow (approved):
Looks good:

on PC
Screenshot from 2022-11-23 20-34-51

on macOS
Screenshot from 2022-11-23 20-34-35

PR #895, opened 2022-11-14
Fix segment list unique key warning

Arnei:
The unique key needs to be on the topmost element, which is now
'ThemedTooltip' instead of div.


lkiesow:
Why does it need to be on the topmost element? What is the observable error that this is fixing?

Arnei:
The observable error is a console warning
```
Warning: Each child in a list should have a unique "key" prop.
```
I have not actually tested if it causes problems in this particular case, but it could cause the state of the app to be faulty when the array in question is changed (element added/removed/swapped etc.). 

The "key" prop needs to be on topmost element in the iteration because React needs it to be (The "key" prop is used by React behind the scenes to track array elements and minimize rerenders I believe).

Issue #858, opened 2022-10-28
Marker for current timeline position hard to notice


lkiesow:
The marker for the current timeline position is hard to notice. In particular, if the current location is the beginning or end of a segment. That are locations where users will often be. The line should at least extend over the waveform.

![Screenshot from 2022-10-28 17-02-11](https://user-images.githubusercontent.com/1008395/198673968-e9658fb0-2b24-4b78-9c96-e029796287fc.png)

PR #906, opened 2022-11-18
Fix timeline marker in subtitle-editor being obscured

Arnei:
Forces the marker to the front via CSS.

Resolves https://github.com/opencast/opencast-editor/issues/858


lkiesow (approved):
That's better:

Screenshot from 2022-11-23 20-07-00

PR #910, opened 2022-11-23
Add editor files to gitignore

lkiesow:
This patch adds a few editor specific files to the `.gitignore` to to prevent them from accidentally being committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,10 @@
 # production
 /build
 
+# editor files
+*.swp
+.vscode/
+
 # misc
 .DS_Store
 .env.local


Arnei (changes requested):
Doesn't seem like it would cause anything relevant to get ignored, so sure why not.

Maybe add a comment as to which editor each line is pertaining to? While `.vscode/` is self-explanatory, I'm not too sure what `*.swp` belongs to (it's vim isn't it?)

lkiesow:
> I'm not too sure what *.swp belongs to (it's vim isn't it?)

Young folk nowadays… :grumble:
When I was young everyone knew that :cane_waving:

lkiesow:
I don't really mind explaining this, but you also didn't really explain any of the other excludes besides the broad categories. Do we need explanations for these?

Arnei (approved):

Arnei:
> I don't really mind explaining this, but you also didn't really explain any of the other excludes besides the broad categories. Do we need explanations for these?

Do we need them? No. Could they maybe potentially be useful in the future? Sure. Do I care enough to block the PR because of it? Absolutely not.

PR #958, opened 2023-01-25
Change subtitle identification from flavor to id

### LINKED ISSUE #896: Detect subtitles language using tags
Following our discussion here https://github.com/opencast/opencast/issues/4407

I changed the way the processing subtitles work from using only the flavor (ex: captions/vtt+de) to flavor & tag approach (F:captions/master, T:lang;de), making a streamlined workflow and less prone to errors.

I haven't found a way in the configuration, to set the subtitles per tag.

Arnei:
*This PR aims at contributing to the implementation of opencast/opencast#4407, but touches on just one of the aspects to realize #4407's vision. It should therefore likely not be merged alone, but alongside other PRs.*

*This PR depends on backend changes in opencast/opencast#4626 !*

Hopefully closes #896?

With subtitles becoming first class citizens in Opencast, their flavor cannot be used as an identifier anymore. This is problematic for the editor, which relies heavily on identifying subtitles by subflavor. This PR switches from subflavor based identification to a track identifier
 based identification.

 The frontend will still try to match subtitle tracks from
 the backend to languages defined in the frontend
 settings. However, it now has to do so by "lang" tag,
 which is not guaranteed to exist. In order to avoid
 creating new subtitle tracks in the backend on every
 save, the frontend settings will therefore require that
 a defined language MUST have a "lang" tag.

 Furthermore, this PR also includes a behavioural
 change in which subtitles are displayed in the
 subtitle select view. Before, only subtitles that were
 specified in the frontend settings were displayed.
With this PR, all subtitles will be displayed, regardless of whether they are specified in the frontens settings or not. This change was made with the rationale that subtitle tags are optional, and subtitle tracks should therefore be editable in the editor even if no "lang" tag is specified.


mliradelc (approved):
I'm not a type script specialist, but seems codewise good. A change necessary for subtitles to work.

PR #959, opened 2023-01-25
Fix language wording ES to local name

cgruber-uos:
We should use local names as language, as it is for "Deutsch" and "English" already. So it must "Español" instead of "Spanish".


Arnei (changes requested):
You might have seen that we have two versions of the config file, this one for development and the other one for release version. The other one is in the root directory of the project, mind changing this there as well?

PR #973, opened 2023-02-01
Improve visibility of input fields

### LINKED ISSUE #932: Subtitle editor, make input fields better visible
Input fields for subtitle text and time marks are nicely formated in dark mode: Input fields have other background colors than the page background and a clear visible border: 

2022-12-07 11_54_15-Greenshot

We should make similar things with the "default" / white skin: 

2022-12-07 11_54_14-Greenshot

Either making the input background light gray or the general background (then for all pages)!

dennis531:
Shows a slightly grey (whitesmoke) background and white input fields in light mode.

This changes should make the input fields better visible, since they are currently displayed in the same background color.

Resolves #932

### DIFF
--- a/src/cssStyles.tsx
+++ b/src/cssStyles.tsx
@@ -386,7 +386,7 @@ export const subtitleSelectStyle = (theme: Theme) => createTheme({
     MuiSelect: {
       styleOverrides: {
         select: {
-          background: `${theme.background}`,
+          background: `${theme.element_bg}`,
           color: `${theme.text} !important`,
         },
         icon: {
--- a/src/main/MainMenu.tsx
+++ b/src/main/MainMenu.tsx
@@ -39,6 +39,7 @@ const MainMenu: React.FC<{}> = () => {
     alignItems: 'center',
     padding: '20px',
     overflow: 'vertical',
+    background: `${theme.menu_background}`,
     ...(flexGapReplacementStyle(30, false)),
   });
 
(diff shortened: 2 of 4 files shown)


Arnei (approved):
Works and looks reasonable to me.

While I'd be fine merging this as is, what do you think about using a different background color for the main menu vs the main content? So for example, the keep `snow` color for the main menu and only set the main content part to `whitesmoke`?

dennis531:
It is a good idea, as it would better separate the menu from the content.

dennis531:
The interface with a white navigation menu looks like this:
image

Should the buttons be displayed in a lighter grey?

Arnei:
> Should the buttons be displayed in a lighter grey?

Hmm, I don't know, what do you think? :D

dennis531:
I would suggest a lighter grey like the following:
image

Arnei:
Seems fine by me. Might wanna do a contrast check with e.g. the WAVE browser add-on to check if the contrast is still high enough (for accessibility reasons).

Arnei:
Seems like the it doesn't make much of a difference. I'm fine merging this either way, I'll let you decide if you want to make more changes.

dennis531:
> Seems fine by me. Can you do a contrast check with e.g. the WAVE browser add-on to check if the contrast is still high enough (for accessibility reasons).

The contrast checker could not detect any constrast issues.

> Seems like the it doesn't make much of a difference. I'm fine merging this either way, I'll let you decide if you want to make more changes.

I added my latter suggestion with the lighter buttons.

Arnei (approved):

Issue #941, opened 2022-12-21
Random strings and wrong special characters in edited subtitle text


Piriya-VM:
We use Vosk to create subtitles and edit them in the subtitle editor.
After saving and publishing, the subtitle file includes strange strings and special character like ä, ü, ö don't get displayed correctly.
See the attached vtt file, screenshot and partial-publish Workflow.
```
WEBVTT

3lPSCXPUvSk3QfzzsrXDj
00:00.420 --> 00:05.930
Apfel

iOmyuq7YoRVkBdHrDHj8m
00:06.930 --> 00:10.530
Hallo ich bin Vosk und versuche nur 
zu Ueberleben irgendwas irgendwas.

3bW92IwxDIrqu0k_HdIeu
00:10.530 --> 00:14.100
orange

zPaQJaT9U0iBY2JI5sqFS
00:14.100 --> 00:18.750
kiwi 
irgendwas irgendwas irgendwas

lBh9Dx0GfstMMqgq-R1XG
00:19.920 --> 00:22.860
ananas
```

![editor_bug](https://user-images.githubusercontent.com/78363047/208886656-158387b7-bee2-4022-9827-49038f1daf93.png)

[partial-publish.txt](https://github.com/opencast/opencast-editor/files/10277064/partial-publish.txt)

DZenker:
I can confirm this bug, this also happens in our installation (OC 12.6).

fsufrre:
I can confirm this bug too. (OC 12.4)
Any ideas when this bug is fixed?

Arnei:
Interestingly, I cannot confirm this bug (OC 0fb1c14, also newer versions). The characters `ä, ü, ö` render normally.

The "strange strings" like `3lPSCXPUvSk3QfzzsrXDj` are subtitle cue ids generated by vosk. I do not know why vosk generates them, but looking at the downloaded subtitle file they are perfectly legal WebVTT. Maybe Paella Player doesn't know how to handle them?

fsufrre:
we are using amberscript. but i think the way of delivery is not the point. while saving the edited subtitle and republish the transcription, the *vtt file gets some extra strings. its visible in paella and also after downloading the *vtt File. I think the magic happens in "saving" process.

Arnei:
True, it seems blaming the strage strings on vosk was hasty on my part.

PR #974, opened 2023-02-02
Add internal identifier for subtitle cues

Arnei:
The subtitle editor requires unique identifiers to work properly. However, the webvtt parser does not provide such ids. WebVTT cue ids are not a good fit either, as they hardly qualifiy as identifiers.

This adds a new field to the subtitle cues to act as a unique identifier. This should also stop the editor from adding randomly generated WebVTT cue ids to subtitle files.

Helps with #941.


Arnei:
@lkiesow You said you wanted to do reviews, how about this one? :D

PR #978, opened 2023-02-17
Omit "preview" in video flavors drop down of subtitle editor

### LINKED ISSUE #927: Subtitle editor, omit "preview" in video flavors drop down
2022-12-06 13_32_19-2022-11_untertitel-editor_v0 6 pptx - PowerPoint

> presentation/preview
> presenter/preview

as value for the drop down menu is confusing. The annex "`/preview`" is same for both options, so no additional information value. It has also bad readability due to no space before or after the slash. 

Let's omit "`/preview`" for both options!

dennis531:
The implementation simply omits the subtypes if subtypes of all tracks are equal

Resolves #927

### DIFF
--- a/src/main/SubtitleVideoArea.tsx
+++ b/src/main/SubtitleVideoArea.tsx
@@ -170,6 +170,11 @@ const VideoSelectDropdown : React.FC<{
 
   // Turn flavor into string
   const stringifyFlavor = (flavor: Flavor) => {
+    // Omit subtype if all flavour subtypes are equal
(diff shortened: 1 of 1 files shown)


Arnei (changes requested):
While this works as advertised, it also breaks the selection part of the dropdown. I am unable to switch to another video. The dropdown updates correctly (in the sense that it displays what I selected), but the video in the player does not change.

Arnei (approved):

Issue #739, opened 2022-07-05
Shortcuts partly broken: Depend on the order you press the keys


lkiesow:
The shortcuts in the editor depend on the order the keys are pressed on the keyboard, which makes them very unreliable and annoying to use. For example, in the cutting view, `Ctrl + Alt + d` does not work if you press these keys in order, but `Alt + Ctrl + d` does work.

Reproducible in Chrome, Chromium and Firefox on Fedora 35 Cinnamon.

Arnei:
Can't reproduce this. Does not matter if I pres `Ctrl` or `Alt` first.

The one thing i found that does not work in Chrome is pressing `Alt + d` before `Ctrl`, because that moves focus to the adress bar.

Works for me in Chrome and Firefox on Fedora 34.

cgruber-uos:
Same result as Arne reports for Win11: Can not reproduce any difference in `Ctrl + Alt + d` vs. `Alt + Ctrl + d` with FF / Edge / Chrome all 106.x. 

Just pressing `Ctrl + d` first leeds to "edit bookmark" dialog and `Alt + d` leads to browser's file menu.

Arnei:
Closing this because given Clemens and my testing, the issue does not seem to be general, but local to Lars.

PR #979, opened 2023-02-17
Additional Hotkey to play back video in the subtitle editor

### LINKED ISSUE #901: Additional Hotkey to play back video in the subtitle editor
Like in other views, the video in the subtitle editor can be played back by pressing `Space`. Unlike other views, the users focus will be inside of an input element most of the time, so pressing `Space` will create a whitespace character instead (as is expected and desired). Therefore, it would be nice to have an alternate hotkey for playing back the video that works while inside an input element. Maybe `Shift` + `Space`.

dennis531:
Currently, users can not play/pause the video in the subtitle input field. These changes solve this problem by introducing an alternate hotkey `Strg` + `Alt` + `Space` to play and pause videos.

Resolves #901

### DIFF
--- a/src/globalKeys.ts
+++ b/src/globalKeys.ts
@@ -42,6 +42,7 @@ export const videoPlayerKeyMap: KeyMap = {
   play: {
     name: "keyboardControls.videoPlayButton",
     sequence: rewriteKeys("Space"),
+    sequences: [rewriteKeys("Control+Alt+Space"), "Space"],
     action: "keydown",
     group: groupVideoPlayer,
   },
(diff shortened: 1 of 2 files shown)


Arnei (approved):

lkiesow:
@Arnei, you approved. I see the effect that pressing `Ctrl`, then `Alt`, then `Space` doesn't work. Pressing `Alt` first, then `Ctrl`, then `Space` works just fine. I see this in both Firefox and Chromium. Not sure if this conflicts with a system-wide hotkey (nothing seems to happen). Does that work for you? If yes, feel free to merge this.

Arnei:
@lkiesow for me both orders (Ctrl --> Alt, Alt --> Ctrl) work. I recall that you described this problem previously in #739 months ago, so I think this is likely a problem on your end (that more users could have anyway, so it may still be worth investigating).

PR #1000, opened 2023-03-17
Video Editor Lock

zismir:
This patch adds functionality to lock the video when in use and depends on https://github.com/opencast/opencast/pull/4658.

### DIFF
--- a/src/main/Body.tsx
+++ b/src/main/Body.tsx
@@ -12,6 +12,7 @@ import { useSelector } from 'react-redux';
 import { selectIsEnd } from '../redux/endSlice'
 import { selectIsError } from "../redux/errorSlice";
 import { settings } from '../config';
+import Lock from "./Lock";
 
 
 const Body: React.FC<{}> = () => {
@@ -39,6 +40,7 @@ const Body: React.FC<{}> = () => {
         <div css={bodyStyle}>
           <MainMenu />
           <MainContent />
+          <Lock />
         </div>
       );
     }
--- a/src/main/Lock.tsx
+++ b/src/main/Lock.tsx
@@ -4,15 +4,8 @@ import { faLock } from "@fortawesome/free-solid-svg-icons";
 import { settings } from "../config";
 import { useDispatch, useSelector } from "react-redux";
 import { setError } from "../redux/errorSlice";
-import { LockData, setLock, video } from "../redux/videoSlice";
+import { setLock, video } from "../redux/videoSlice";
 import { client } from "../util/client";
-
-export type ILock = {
-  lockingActive: boolean;
-  lockRefresh: number;
-  lockState: boolean;
-  lock: LockData;
-};
 
 const Lock: React.FC<{}> = () => {
   const [state, setState] = useState({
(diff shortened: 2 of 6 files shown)


src/redux/__tests__/videoSlice.test.ts
```diff
@@ -133,8 +133,8 @@ describe('Video reducer', () => {
 
     // Assert
     const rootState = { videoState: nextState };
-    expect(selectCurrentlyAt(rootState)).toEqual(20);
-    expect(selectActiveSegmentIndex(rootState)).toEqual(1);
+    expect(selectCurrentlyAt(rootState)).toEqual(21);
```
Arnei, line 136: Locking should not change the layout at all. Instead of changing the test, change your implementation please.

src/main/Body.tsx
```diff
@@ -39,6 +40,7 @@ const Body: React.FC<{}> = () => {
         <div css={bodyStyle}>
           <MainMenu />
           <MainContent />
+          <Lock />
```
Arnei, line 43: Since this doesn't render anything, it could probably go even higher in the hierarchy.

src/main/Lock.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export type ILock = {
+  lockingActive: boolean;
+  lockRefresh: number;
+  lockState: boolean;
+  lock: LockData;
+};
+
+const Lock: React.FC<{}> = () => {
+  const [state, setState] = useState({
```
Arnei, line 18: Also goes unused, since all these values are stored in redux. Respectively, all "setState" calls are superfluous.

Arnei (changes requested):
Seems like a generally reasonable approach, but does not work:

- Does not check whether locking is enabled in the backend. I think either you or I are misunderstanding what "lockingActive" is supposed to mean.
- Does not refresh the lock

src/redux/videoSlice.ts
```diff
@@ -212,6 +230,11 @@ const videoSlice = createSlice({
         state.originalThumbnails = state.tracks.map((track: Track) => { return {id: track.id, uri: track.thumbnailUri} })
 
         state.aspectRatios = new Array(state.videoCount)
+        state.lockingActive = action.payload.locking_active
+        state.lockRefresh = action.payload.lock_refresh
+        state.lockState = action.payload.lock_state;
```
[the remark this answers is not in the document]
JamesUoM: Code in in review: https://github.com/opencast/opencast/pull/4658

src/redux/videoSlice.ts
```diff
@@ -28,6 +28,16 @@ export interface video {
   title: string,
   presenters: string[],
   workflows: Workflow[],
+
+  lockingActive: boolean,
+  lockRefresh: number | 0,
```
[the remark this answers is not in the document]
zismir: I have removed 0.

src/redux/__tests__/videoSlice.test.ts
zismir: I have reverted the unit test.

src/main/Lock.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { useCallback, useEffect, useState } from "react";
+import { faLock } from "@fortawesome/free-solid-svg-icons";
+
+import { settings } from "../config";
+import { useDispatch, useSelector } from "react-redux";
+import { setError } from "../redux/errorSlice";
+import { LockData, setLock, video } from "../redux/videoSlice";
+import { client } from "../util/client";
+
+export type ILock = {
```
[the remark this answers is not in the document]
zismir: I have removed `ILock` type.

src/redux/__tests__/videoSlice.test.ts
Arnei: Thanks to #1053 I understand now that you were just fixing a test that was already failing. Good idea to submit it as a seperate PR.

JamesUoM:
This needs to be submitted from the UoM fork. Will close and resubmit.

Issue #997, opened 2023-03-02
Order workflows by displayOrder


JamesUoM:
Use the provided displayOrder value to sort the workflows.

```
"workflows":[{
  "id":"publish",
  "name":"Publish",
  "displayOrder":1000,
"description":"Process media files and publish them."
},{
  "id":"manchester-after-editing",
  "name":"Publish to Video Portal",
  "displayOrder":9060,
  "description":"Restore workflow properties, compose and publish"
}]```

PR #1005, opened 2023-03-21
Workflow Filter

### LINKED ISSUE #36: restrict workflows available in the editor
There may be more editor workflows in the system than you would like to provide to a regular user (lecturer). 
Based on the context there should be a way to restrict the workflows available in the "select workflow" section. 

One idea could be to provide the available workflows with a GET-parameter.

zismir:
This patch adds functionality to filter workflows.

Fixes #36 and #997.

### DIFF
--- a/.github/mock/editor/ID-dual-stream-demo/edit.json
+++ b/.github/mock/editor/ID-dual-stream-demo/edit.json
@@ -5,7 +5,13 @@
       "id": "publish",
       "name": "Publish",
       "displayOrder": 1000,
-      "description": ""
+      "description": "Process media files and publish them."
+    },
+    {
+      "id":"manchester-after-editing",
+      "name":"Publish to Video Portal",
+      "displayOrder":9060,
+      "description":"Restore workflow properties, compose and publish"
     }
   ],
   "tracks": [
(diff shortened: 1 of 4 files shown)


lkiesow:
Hi @zismir, 
thanks for your contribution.

I'm not quite sure what you are trying to do here. I think the issue from @rrolf was asking for an admin to filter the list of workflows presented to users down more than with just the `editor` tag. Since the removal of the old editor is already planned, that should be a mood point since then, the tag is used for this editor only and you can present only the workflows you want.

Using an additional drop-down to filter by arbitrary numbers which aren't even exposed to users in addition is not a good idea. As a user, I don't know about display order and therefor, I cannot operate the drop-down in a sensible way to filter the workflows in a useful way for me.

More than that, the numbers seem arbitrary and by default, I get a dropdown, which doesn't seem to do anything. It also breaks the idea to hide the workflow selection if there is no selection (if there is but one workflow) and it also doesn't really work in dark mode:

Screenshot from 2023-03-31 19-34-10

That said, ordering by `displayOrder` does seem to work. If you submit a separate patch for that, I can probably review and merge that quickly.


Some general remarks/requests about this. It's really really hard as a reviewer to guess intentions by code. In the future, it would be awesome if you could briefly explain what you want to do, how you want to do that and why. It's much easier to understand a patch if you know the intention. Here I just have the code and am confused about your intention. Also, while it's good to reference issues, please don't use them as an excuse to not write these things down. Issues often contain vague ideas and discussions and I don't know which ideas you took or discarded when working on the pull request.

Second, please try submitting one patch for one change. The ordering seems good and straightforward for me. If you had submitted that in a separate patch, I might have already merged that. It also makes reviews much simpler since you don't constantly have to wonder what idea/feature/bugfix any change belongs to.

lkiesow:
This also seems to crash the editor if no workflows are tagged editor. See #1011

JamesUoM:
This pull request was made by mistake and should have been pulled.

PR #1012, opened 2023-03-31
Add ICLA Check

lkiesow:
This patch adds the ICLA check used in Opencast's main repository. The editor follows the same rules and regulations and contributors should file an ICLA.


Arnei (approved):

lkiesow:
It's approved and went through the technical meeting → merging

PR #1028, opened 2023-04-05
No Workflow Error

lkiesow:
This patch updates the error message users get if there are no workflows usable by the editor. It sounded a bit weird and I didn't quite agree with the grammar :D


Arnei (approved):
Sure, seems reasonable

PR #1033, opened 2023-04-13
Basic zoom functionality for the cutting view

### LINKED ISSUE #107: Make the timeline "zoomable"

Arnei:
Based off of the subtitle view. Implements a very basic zoom functionality, hoping that it's better than nothing. Includes:
- A Slider to control the zoom level
- A horizontal scrollbar to move the timeline
- Drag n' Drop to move the timeline

Known issues:
- ~The scrubber/playhead is cut off if placed either at 0 or at the maximum duration. I might have css'd myself into a corner here.~

Possible future feature extensions:
- Hotkeys
- Centering the zoom on the scrubber/playhead while zooming in (difficult to do without jittering)
- Configuration

Resolves #107.


ziegenberg:
@Arnei could you rebase and fix the conflicts? I'd like to test this.

ziegenberg:
Thank you @Arnei! :)

ziegenberg:
Works as described. I only have one small visual bug to report. The slider should probably live in its own div like the other controls around it. At the moment it almost overlaps the control to the left of it. See the screenshot for reference.

image

ziegenberg:
Works now (mostly). 🎉 

I found some funny bugs, though:

- WHEN I scroll/move the timeline,
THEN the scrubber does not move together with it.
- WHEN I zoom the timeline
AND the scrubber moves out of the viewport
AND I scroll/move the timeline afterwards,
THEN the scrubber stays gone, even if it should be back in the viewport.
- WHEN I zoom the timeline
AND I then scroll/move the timeline a bit so that the scrubber still remains in the viewport
AND then click on the timeline, 
THEN the scrubber moves with a certain offset.

ziegenberg:
Looks better now!

1. Scrolling the timeline also moves the scrubber ✅
2. Zooming the timeline with the scrubber going off-screen does not make it disappear when scrolling it back into the viewport ✅
3. The scrubber gets placed correctly when zooming and then scrolling the timeline. ✅

Features nice to have:
- Centering the zoom on the scrubber/play-head while zooming in
- Scrolling the timeline with the mouse wheel while being zoomed in
- Having hotkeys for zooming in/out and scrolling left/right while being zoomed in

Arnei:
> Scrolling the timeline with the mouse wheel while being zoomed in

You can do this using `Shift` + mouse wheel (courtesy of the `react-indiana-drag-scroll` library, which I'm inclined to rely on, as properly implementing this is a pain). Not sure how to inform users about that though.

Arnei:
> Centering the zoom on the scrubber/play-head while zooming in

I would skip this for now, seems to be more difficult to get to work properly than I anticipated.

ziegenberg:
> > Scrolling the timeline with the mouse wheel while being zoomed in
> 
> You can do this using `Shift` + mouse wheel (courtesy of the `react-indiana-drag-scroll` library, which I'm inclined to rely on, as properly implementing this is a pain). Not sure how to inform users about that though.

How about a pop over when hovering on the timeline?

ziegenberg:
> > Centering the zoom on the scrubber/play-head while zooming in
> 
> I would skip this for now, seems to be more difficult to get to work properly than I anticipated.

I think that's fair, but we should add an issue to not forget about that. It's a rather big UX annoyance / it's un-intuitive that it doesn't zoom centered on the scrubber.

Arnei:
> > > Centering the zoom on the scrubber/play-head while zooming in
> > 
> > 
> > I would skip this for now, seems to be more difficult to get to work properly than I anticipated.
> 
> I think that's fair, but we should add an issue to not forget about that. It's a rather big UX annoyance / it's un-intuitive that it doesn't zoom centered on the scrubber.

Definitely. I'd add the an issue for that after this is merged, creating an issue for a feature that is not even merged yet feels weird.

kai-wng-uos:
Also, it would be nice if you could zoom using the scroll wheel. In that case, the zoom would focus on the cursor. This is only a nice to have feature, and I guess it could be confusing if you just accidentally touch the scroll whel and suddenly you're zoomed in. But maybe someone has a good idea on how to implement such a function.

JulianKniephoff:
> But maybe someone has a good idea on how to implement such a function.

Maybe use Alt+Scroll, like we already do in Paella? (Which in turn is inspired by Google Maps, IIRC.) Try scrolling while hovering over the video here for example: https://tobira.opencast.org/v/Ijpj-vfg1TH

Would that be acceptable? 🤔

dagraf:
> Currently, when using the zoom, you automatically zoom in on the beginning of the waveform. This is highly confusing, the zoom should focus on the current position of the postion marker, so you automatically zoom in to where you currently are in the video.

I strongly agree to focus on the current position of the playhead when zooming. The current behavior is particularly confusing when I zoom in or out again after having zoomed in and having moved the horizontal scrollbar (e.g. to the position of the playhead). I actually like the behavior of the old editor. There the zoomed-in section also gets moved when the playhead would run out of it. That would be nice to have also in the new editor, too.

KatrinIhler (changes requested):
See improvements discussed above.

luniki:
I have fixed the zoom center problem in https://github.com/luniki/opencast-editor/commits/bugfix/zoom-center/

The zooming logic is like this:

If the scrubber is visible just before zooming, the timeline centers the scrubber while zooming.
If the scrubber is not visible just before zooming, the timeline zooms to the centers of the currently visible timeline.

luniki:
Weiter oben wurde als Nice-to-have genannt, dass auch mit dem Mausrad gezoomt werden kann. Soll das auch noch ergänzt werden?

Arnei:
Applied lunikis patch.

snoesberger:
> I have fixed the zoom center problem in https://github.com/luniki/opencast-editor/commits/bugfix/zoom-center/

Looks good to me, works as you described.

But I've found a few small issues:
- When I've zoomed in on the timeline and want to drag and drop a cut-marker, I can only move the cut-marker a few pixels. Then the whole timeline moves.
- The zoom tooltip is missing.
- I'd prefer a larger zoom range. We sometimes have recordings for special events that are up to 4h long. With full zoom, it is still difficult to precisely cut such a video.

Arnei:
Already told this luniki in private chat, but for posterities sake: When changing the zoom level, the playhead jumps (jitters?) back and forth. This seems to be a purely visual issue:

[Bildschirmaufzeichnung vom 2024-05-17, 14-09-59.webm](https://github.com/opencast/opencast-editor/assets/14070005/2a151202-00b6-4dfe-afb1-d900df9a1dd5)

luniki:
Thanks for your comments.

>     * When I've zoomed in on the timeline and want to drag and drop a cut-marker, I can only move the cut-marker a few pixels. Then the whole timeline moves.

This is quite difficult! It almost works in Firefox, but it's so slow in Chromium. The problem is the constant updating of the `currentlyAt` position while dragging, which updates the currently dragged elements. Removing the `onControlledDrag` handler makes it usable again, but we lose the UI feedback of the changed scrubber position.

I have debounced the `onControlledDrag` handler, which delays calling the handler until 200 milliseconds have passed since the last time the handler was called. 

IMHO this is a good compromise: 
[0002-Debounce-drag-handler-of-the-scrubber.patch.txt](https://github.com/opencast/opencast-editor/files/15386289/0002-Debounce-drag-handler-of-the-scrubber.patch.txt)

>     * The zoom tooltip is missing.

I have added the tooltip: 
[0001-Add-zoom-tooltip.patch.txt](https://github.com/opencast/opencast-editor/files/15386150/0001-Add-zoom-tooltip.patch.txt)


>     * I'd prefer a larger zoom range. We sometimes have recordings for special events that are up to 4h long. With full zoom, it is still difficult to precisely cut such a video.

What is the maximum zoom factor we should support?

luniki:
> I have fixed the zoom center problem in https://github.com/luniki/opencast-editor/commits/bugfix/zoom-center/
> 
> The zooming logic is like this:
> 
> If the scrubber is visible just before zooming, the timeline centers the scrubber while zooming. If the scrubber is not visible just before zooming, the timeline zooms to the centers of the currently visible timeline.

There is a bug in computing the zoom center. This patch fixes it: 
[0003-Fix-fast-zooming-bug.patch.txt](https://github.com/opencast/opencast-editor/files/15386317/0003-Fix-fast-zooming-bug.patch.txt)

luniki:
> Already told this luniki in private chat, but for posterities sake: When changing the zoom level, the playhead jumps (jitters?) back and forth. This seems to be a purely visual issue:
> Bildschirmaufzeichnung.vom.2024-05-17.14-09-59.webm

The jumping of the scrubber is a result of the asynchronous update of:
1. `scrollLeft` of the `ScrollContainer`
2. `width` of the `div` that contains the `Waveforms`
3. `controlledPosition` of the `Scrubber`

When we zoom in, the `scrollLeft` updates before the `width` and the `controlledPosition`.

I do not have a clue on how to synchronize all of this.

snoesberger:
> What is the maximum zoom factor we should support?

I'd prefer not to have a fixed zoom factor, but to always have the same time span, e.g. 20 sec., in the view when the zoom is at maximum (regardless of the video duration). Zoom slider on far left: full waveform in view, zoom slider on far right: always 20 sec in view.

It would also be nice if the slider was wider to give more precise control over the zoom level.

luniki:
If you set the maximum zoom factor in this way, the width of the browser will play a role. Which lead me to an interesting bug in the current implementation: If you change the window width, the timeline shifts in an unexpected way.

luniki:
> It would also be nice if the slider was wider to give more precise control over the zoom level.
Fixed: 
[0004-Increase-the-width-of-the-zoom-slider.patch.txt](https://github.com/opencast/opencast-editor/files/15397989/0004-Increase-the-width-of-the-zoom-slider.patch.txt)

luniki:
> > Already told this luniki in private chat, but for posterities sake: When changing the zoom level, the playhead jumps (jitters?) back and forth. This seems to be a purely visual issue:
> > Bildschirmaufzeichnung.vom.2024-05-17.14-09-59.webm
> 
> The jumping of the scrubber is a result of the asynchronous update of:
> 
>     1. `scrollLeft` of the `ScrollContainer`
> 
>     2. `width` of the `div` that contains the `Waveforms`
> 
>     3. `controlledPosition` of the `Scrubber`
> 
> 
> When we zoom in, the `scrollLeft` updates before the `width` and the `controlledPosition`.
> 
> I do not have a clue on how to synchronize all of this.

I will try to position the scrubber by using a percentage instead of a changing px value.

luniki:
Nope, this does not work. The only way I can imagine removing the flickering of the scrubber is separating the elements of the scrollable waveform and the scrubber. The scrubber would have to be positioned separately in a different element.

snoesberger:
> If you set the maximum zoom factor in this way, the width of the browser will play a role. Which lead me to an interesting bug in the current implementation: If you change the window width, the timeline shifts in an unexpected way.

@luniki is this a new bug that we already have now or will this bug be added by changing the way we set the max zoom factor? And what do you mean by "shifts in an unexpected way"?

snoesberger:
I have tested the latest changes. Here are the results of my tests for the various open issues:

- [ ] **Drag and drop of cut marker when zoomed in**
Does still not work for me.

- [ ] **Always 20 sec of waveform in view when max. zoom**
Still 10x max zoom factor.

- [x] **Increased width of zoom slider**
implemented

- [ ] **jumping of the scrubber when zooming**
still the same behavior

- [x] **Zoom slider tooltip**
implemented

oas777:
Late to the game with a question: Do the shortcuts work for you? Both me and my colleague failed to activate them.

snoesberger:
> Late to the game with a question: Do the shortcuts work for you? Both me and my colleague failed to activate them.

The shortcuts for zooming in and out (Shift + Option + z/t or "+"/"-") don't work for me either. However, if I first activate the zoom slider by clicking on it, I can use the "left"/"right" keys to change the zoom level.

Arnei:
I've tested `Shift + Alt/Option + z/t` with the latest Chrome, Firefox and Safari, and it worked in all three.  `+/-` did not work in Safari, but is also not expected to work in every browser, which is the reason for the alternative shortcut.

Any further information like Browser version, System or detailed description/video of what you are doing would be appreciated.

oas777:
My observations for Windows 11 with Firefox and Chrome:

- SHIFT+ALT+z seems to work "once", but if you interact with UI afterwards, the hotkey doesn't work anymore.
- SHIFT+ALT+t seems to trigger a screen recording here, but maybe that's my Lenovo or the Snipping tool interfering.
- Also, I find the syntax of the hotkey description confusing, what do the additional + and - stand for?
grafik

snoesberger:
> Any further information like Browser version, System or detailed description/video of what you are doing would be appreciated.

I tested it on a MacBook Pro (Sonoma 14.5) and Shift + Option + z/t or "+"/"-" didn't work at all in Firefox, Safari, Chrome and Opera Browsers ().

oas777:
Oh, so that's an alternative to z/t? They don't work at all at my end.

luniki:
> I have tested the latest changes. Here are the results of my tests for the various open issues:
> 
>     * [ ]  **Drag and drop of cut marker when zoomed in**
>       Does still not work for me.
> 
>     * [ ]  **Always 20 sec of waveform in view when max. zoom**
>       Still 10x max zoom factor.
> 
>     * [x]  **Increased width of zoom slider**
>       implemented
> 
>     * [ ]  **jumping of the scrubber when zooming**
>       still the same behavior
> 
>     * [x]  **Zoom slider tooltip**
>       implemented

I have fixed the drag and drop of the cut marks and changed the timeline zoom maximum depending on the video length in these patches:

[0002-Enable-drag-and-drop-of-cut-markers-when-zoomed-in.patch.txt](https://github.com/user-attachments/files/15759672/0002-Enable-drag-and-drop-of-cut-markers-when-zoomed-in.patch.txt)
[0001-Compute-maximum-zoom-factor-depending-on-the-length-.patch.txt](https://github.com/user-attachments/files/15759673/0001-Compute-maximum-zoom-factor-depending-on-the-length-.patch.txt)

I cannot find a solution for the jumping of the scrubber. I would like to work on a proof of concept for using another library specialized in this. For a first version I would estimate about 2-3h to get an idea on how (and if) this could work.

Arnei:
> * [x]  **Drag and drop of cut marker when zoomed in**
  Does still not work for me.
>* [x]  **Always 20 sec of waveform in view when max. zoom**
  Still 10x max zoom factor.
>* [x]  **Increased width of zoom slider**
  implemented
>* [ ]  **jumping of the scrubber when zooming**
  still the same behavior
>* [x]  **Zoom slider tooltip**
  implemented

Arnei:
Also regarding the hotkey issue. Is it only the zoom hotkeys that are not working, or are you having this problem with other hotkeys as well? If it is the latter, I suggest we discuss it in a separate issue.

snoesberger:
> As already elaborated, fixing "jumping of the scrubber when zooming" will require a lot more work and as such we should consider if want to merge this even with this bug present.

We can live with this behaviour for now. I would suggest that we open a separate issue for this cosmetic problem.

> 
> Also if you try to test "Always 20 sec of waveform in view when max. zoom" with the default "Dual Stream Demo" video, you will notice that the max zoom will not be at 20 seconds. That is because the video is relatively short with only about a minute of playtime, for which 20 seconds don't really make sense. For longer videos "Always 20 sec of waveform in view when max. zoom" will work.

I agree with you, it makes sense and is fine for us.

snoesberger:
> Also regarding the hotkey issue. Is it only the zoom hotkeys that are not working, or are you having this problem with other hotkeys as well? If it is the latter, I suggest we discuss it in a separate issue.

Some of the shortcuts for the timeline don't work for me either (Shift + Option + i, Shift + Option + k, Up, Down). So, I'll create a separate issue for that.

src/redux/videoSlice.ts
```diff
@@ -489,6 +503,14 @@ const setThumbnailHelper = (state: video, id: Track["id"], uri: Track["thumbnail
   }
 };
 
+const ZOOM_SECONDS_VISIBLE = 20 * 1000;
```
jduehring, line 506: For videos of 20 seconds or less, I can not use the zoom-slider at all because the returned value is used as maximum and will always be 1. This might be fine since the zoom level is already very detailed on a short video, but it would probably make sense to have some kind of visual indication / explanation as to why the user can't drag the slider. I was a bit confused at first.

jduehring (commented):
Other than my first comment, looks  good to me and works as expected. I experienced the "jumping of the scrubber when zooming" but I see this has already been addressed.

src/redux/videoSlice.ts
jduehring: Maybe even just hide the zoom bar completely for short videos < 20sec?

luniki: Hiding is nice. On the other hand, we could always have a minimal "maximum zoom factor" of 2.

snoesberger: Personally, I'd prefer to have a minimum zoom factor. That way the user always has the same experience and never has to wonder where the zoom slider has gone.

luniki: I have changed the minimal maxZoom factor. This patch takes care of it: 
[0001-Update-the-minimal-maxZoom-from-1-to-2.patch.txt](https://github.com/user-attachments/files/15884514/0001-Update-the-minimal-maxZoom-from-1-to-2.patch.txt)

jduehring: @luniki Now the zoom bar seems to be not correctly shown for short videos: 

image

Tested with "ID-3d-print" video.

jduehring: @luniki Now the zoom-bar seems to be not correctly shown for short videos. I guess the buttons are not intended? 

image

Tested with the `ID-3d-print` video on develop.

snoesberger: @jduehring I haven't noticed your problems on https://test.editor.opencast.org/1033/2024-06-18_10-33-57/?id=ID-3d-print. My PC displays the zoom slider as it should (tested with Safari, Firefox, Chrome and Opera).

jduehring: Hmm strange, might have been a local hiccup on my side. Seems to work now. So all in all I have no more complaints and its ready to merge from my side.

PR #1035, opened 2023-04-14
Make Studio and Editor look alike

lkiesow:
Opencast Studio and the Opencast Editor are some of the most prominent new tools of Opencast. It would be nice if the look and feel of both would be more similar so that it feels like they belong to the same service.

This patch is a first step. It introduces a header and logo similar to Studio.

In the future, we can also use this to give users easy access to language options and other settings like with Studio. But that is something for future patches.

Screenshot from 2023-04-14 12-47-15


Arnei (changes requested):
This introduces a permanent vertical scrollbar into the application. Could you get rid of that?

Arnei:
Also we probably want a new class for the Header, rather than put everything into `App.tsx`. But that will probably happen organically when the Header is developed further, so whatever.

Also maybe the header should look different in high contrast mode, but I don't know about that. @naweidner , what do you think?

Looks good otherwise. Finally, a header!

lkiesow:
- [x] Moved header code to its own file
- [x] Adjusted the page height to avoid scroll bars
- [x] Now avoiding gray in high contrast mode

src/main/Timeline.tsx
```diff
@@ -42,7 +42,7 @@ const Timeline: React.FC<{
   setCurrentlyAt: ActionCreatorWithPayload<number, string>,
   setIsPlaying: ActionCreatorWithPayload<boolean, string>,
 }> = ({
-  timelineHeight = 250,
+  timelineHeight = 200,
```
Arnei, line 45: Changes look good to me!

One question though, why change the height of the timeline? To compensate for the header? For the resolutions I checked, the addition of the header does cause a scrollbar to appear for resolutions where there previously were none. Maybe I missed one? Or do you just like the look better? :D

naweidner:
> Also maybe the header should look different in high contrast mode, but I don't know about that. @naweidner , what do you think?

Yes, I agree. It has been implemented now and think it also looks better in black than grey.


> In the future, we can also use this to give users easy access to language options and other settings like with Studio

Could a border at the bottom of the header make sense (only in high contrast mode)? 
I'm already thinking further when there are more elements in the header, such as language selection and other options, as mentioned above. So that there is a visible separation between header and content.

src/main/Timeline.tsx
lkiesow: On a wide-screen monitor, the current code already causes scrollbars to appear because while the overall height is calculated based on the window height, the components in the cutting view seem to have a height calculated so that things just fit in Firefox and Chrome when using a 16:9 aspect ratio.

You can see this when your window is wider than 16:9 if you look at where the menu end on the left side (and you get a scrollbar). Here is a screenshot from editor.opencast.org:

Screenshot from 2023-04-17 19-54-25

With my patch the common 16:9 resolution now already causes this since there is slightly less vertical space. Adjusting the waveform height takes that into account, restoring the best guess approach from before.

When you complained about the scrollbar, the overall height was also too large. That is fixed properly.

lkiesow:
> Could a border at the bottom of the header make sense (only in high contrast mode)?

Done:

Screenshot from 2023-04-17 20-12-15

src/main/Timeline.tsx
Arnei: I'll admit, I did not check wider than 16:9 resolutions (because I keep thinking this screen is 16:10 while it is actually 16:9...). In that case changing the timeline height seems like a reasonable fix.

In order to avoid this problem in the future, we probably want to introduce a maximum height for the video area at some point. Since the video area always consumes the maximum amount of available width while keeping the video aspect ratios. But that's for another PR.

Arnei (approved):

PR #1036, opened 2023-04-18
`npm audit fix`

JulianKniephoff:
Just what it says in the title. :shrug:


Arnei (approved):
Does do what it says in the title, which seems to not have broken anything.

PR #1053, opened 2023-05-10
Updated tests that were failing

zismir:
This patch updates tests that were failing.


Arnei (approved):
Thanks for fixing this!

PR #1055, opened 2023-05-11
Drop unused dependencies

Arnei:
Cleans up the package.json a bit by dropping unused dependencies.

An interesting side effect of this is that we now use a more modern DateTimePicker in the Metadata tab. I believe it provides a vastly better experience than the old one, so I just happily accepted this. Please do give it a try.


package.json
```diff
@@ -54,13 +45,10 @@
     "react-resizable": "^3.0.5",
     "react-scripts": "5.0.1",
     "react-select": "^5.7.2",
-    "react-use": "^17.4.0",
-    "react-virtualized-auto-sizer": "^1.0.15",
+    "react-virtualized-auto-sizer": "^1.0.6",
```
lkiesow, line 48: Is this downgrade intentional?

PR #1056, opened 2023-05-12
Update to React 18

Arnei:
*Includes #1055, as some of the dependencies that it removes are not compatible with React 18.*

Bumps React version to 18, and includes all the necessary changes that React 18 demands.

Does not yet include the switch to the new rendering API that comes with React 18. This is just about getting everything to work after bumping the version number.


src/main/WorkflowSelection.tsx
```diff
@@ -182,15 +181,26 @@ const WorkflowButton: React.FC<{stateName: string, workflowId: string, workflowD
...
+
+  return (
+    <Radio
+      color="default"
+      css={style}
+      checked={false}
+    />
+  )
+}
+
```
LukasKalbertodt, line 204: Something doesn't work here. When running this locally, I cannot select any workflow. At least I don't get any visual feedback and looks like none is selected:

image

LukasKalbertodt (changes requested):

src/main/SubtitleVideoArea.tsx
```diff
@@ -183,50 +181,37 @@ const VideoSelectDropdown : React.FC<{
...
-              label={t("subtitleVideoArea.selectVideoLabel") ?? undefined}
-              name={dropdownName}
-              data={selectData()}
-            />
-          </ThemeProvider>
-
-          <OnChange name={dropdownName}>
-            {(value, previous) => {
+    <>
+      <div>{t("subtitleVideoArea.selectVideoLabel") ?? undefined}</div>
```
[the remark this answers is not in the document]
Arnei: There were some weird type issues in the past. Good to hear that these can go.

src/main/WorkflowSelection.tsx
```diff
@@ -71,7 +70,7 @@ const WorkflowSelection : React.FC<{}> = () => {
   };
 
   // Layout template
-  const render = (topTitle: string, topText: {} | null | undefined, hasWorkflowButtons: boolean,
+  const render = (topTitle: string, topText: JSX.Element | null | undefined, hasWorkflowButtons: boolean,
```
LukasKalbertodt, line 73: Not important but I simply noticed: it's probably fine to remove the `null | undefined` as the function is just called at three places, each time passing a non-null value.

```suggestion
  const render = (topTitle: string, topText: JSX.Element, hasWorkflowButtons: boolean,
```

src/main/KeyboardControls.tsx
```diff
@@ -27,7 +27,7 @@ const Group: React.FC<{name: TFuncKey, entries: KeyMapDisplayOptions[]}> = ({nam
 
   return (
     <div css={groupStyle}>
-      <h3 css={headingStyle}>{t(name)}</h3>
+      <h3 css={headingStyle}>{t(name) as TFuncKey}</h3>
```
LukasKalbertodt, line 30: This is unfortunately not the proper solution for this problem. What's going on here is that `name` is of type `TFuncKey` as specified in the function signature. That type contains all valid strings to index the translation file. Unfortunately, this includes "prefixes". So say your translation file is:

```json
{
    "foo": {
        "bar": "Hello",
        "baz": "Bye"
    }
}
```

Then obviously, `foo.bar` and `foo.baz` are valid `TFuncKey`. But so is `foo`. Calling `t("foo")` will return the object `{ bar: "Hello", baz: "Bye" }`. So without knowing what specific key `t` is called with, only knowing that it is a `TFuncKey`, Typescript has to assume that `t` returns a string OR an object. That's why typescript complains: you can't render an object (pass an object as children in JSX). And that is something that changed with React 18, so that's why it occurs now.

It would be really nifty to have an alternative type to `TFuncKey` that describes all keys that, when given to `t`, will return a string. I don't think such a type exists, unfortunately. There are various ways how to derive such a type, but the likely easier way here is to just define a type with those few fixed options that are expected here. These are the four constants in `globalKeys.ts`. Unfortunately, all of this would be a bit involved as you would need to carry that type through several layers of `globalKeys.ts` and `KeyboardControls.tsx`. Or you can of course just use an `as` cast, as is already done.

Why is this specific solution not ideal? Well, it just casts the output of `t` to `TFuncKey` which is just a subset of `string`. Since all strings are renderable in JSX, this is too. But this is not correct in the sense that `t` will absolutely return strings that are not a `TFuncKey`. I mean... this is "just" wrong on the type level, right now this won't crash or do anything wrong at runtime.

Sorry for the long text, but I hope that explained a bit. I'll leave the decision to you what to do. Not going with the involved solution is fine. But at least the cast should be `as string` then to take confusion out of this. This would actually make sense as we say "we know that what's returned here is always a string, not an object".

Arnei: Thanks for the detailed explanation.

> It would be really nifty to have an alternative type to TFuncKey that describes all keys that, when given to t, will return a string. I don't think such a type exists, unfortunately. There are various ways how to derive such a type, but the likely easier way here is to just define a type with those few fixed options that are expected here. These are the four constants in globalKeys.ts. Unfortunately, all of this would be a bit involved as you would need to carry that type through several layers of globalKeys.ts and KeyboardControls.tsx. Or you can of course just use an as cast, as is already done.

That sounds like it would really bloat the code for minimal gain, especially if there are a lot of possibly valid values. I'll say casting to string is good enough for now.

LukasKalbertodt (approved):
All comments resolved and the workflow selection works now!

PR #1057, opened 2023-05-15
Update to react 18 rendering pipeline

Arnei:
React 18 brings with it a new and improved rendering method. React reminds everyone of this in the console:

> ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it’s running React 17. Learn more: https://reactjs.org/link/switch-to-createroot

This PR updates the editor to that new method, but has not yet fixed all the errors it causes.
Known errors:
- ~All of our global hotkeys do not work anymore, except the timeline ones.~
- ~The list component in the subtitle editor now causes an endless loop, completely breaking the editor.~


Arnei:
Obvious show stoppers have been addressed, so this is finally ready for review.

src/main/CuttingActions.tsx
```diff
@@ -36,9 +36,7 @@ const CuttingActions: React.FC = () => {
    * @param action redux event to dispatch
    * @param ref Pass a reference if the clicked element should lose focus
    */
-  const dispatchAction = (event: KeyboardEvent | SyntheticEvent, action: ActionCreatorWithoutPayload<string>, ref: React.RefObject<HTMLDivElement> | undefined) => {
-    event.preventDefault()                      // Prevent page scrolling due to Space bar press
-    event.stopPropagation()                     // Prevent video playback due to Space bar press
+  const dispatchAction = (action: ActionCreatorWithoutPayload<string>, ref: React.RefObject<HTMLDivElement> | undefined) => {
```
LukasKalbertodt, line 39: No need to change but just a hint:

```suggestion
  const dispatchAction = (action: ActionCreatorWithoutPayload<string>, ref?: React.RefObject<HTMLDivElement>) => {
```

This makes the parameter optional, which means you could call `dispatchAction(cut)` below, for example.

src/main/KeyboardControls.tsx
```diff
@@ -95,20 +88,18 @@ const Entry: React.FC<{params: KeyMapDisplayOptions}> = ({params}) => {
...
-          </div>
-        ))}
-      </div>
+      <div css={labelStyle}><Trans>{name || t("keyboardControls.missingLabel")}</Trans></div>
+      {sequences.map((sequence, index, arr) => (
+        <div css={sequenceStyle} key={index}>
+          {sequence.map((singleKey, index) => (
+            <>
+              <div css={singleKeyStyle} key={index}>{singleKey}</div>
+              {sequence.length - 1 !== index ? <div css={orStyle}>+</div> : ''}
```
LukasKalbertodt, line 97:
```suggestion
              {sequence.length - 1 !== index && <div css={orStyle}>+</div>}
```

That works too. And if you move that line above the single key div, you can just check `index > 0`.

src/main/KeyboardControls.tsx
```diff
@@ -130,30 +119,22 @@ const KeyboardControls: React.FC = () => {
...
 
       const groups: JSX.Element[] = [];
-      for (const key in obj) {
-        if (obj[key].length > 0) {
-          groups.push(<Group name={key as ParseKeys} entries={obj[key]} key={key}/>);
-        }
-      }
+      Object.entries(KEYMAP).forEach(([key, value], index) => {
+        const entries : { [key: string]: string[][] } = {}
+        Object.entries(value).forEach(([, value]) => {
```
LukasKalbertodt, line 127:
```suggestion
        Object.values(value).forEach((value) => {
```

Also, `key` and `value` (both, the outer and inner `value`) should probably get more descriptive names. It's currently very generic and without knowing the exact structure of `KEYMAP` one has no idea what these variables contain.

src/main/KeyboardControls.tsx
```diff
@@ -130,30 +119,22 @@ const KeyboardControls: React.FC = () => {
...
+          const sequences = value.key.split(",").map(item => item.trim())
+          const lol: string[][] = []
+          Object.entries(sequences).forEach(([, value]) => {
+            const keys = value.split("+").map(item => item.trim())
+            lol.push(keys)
+          })
+          entries[value.name] = lol
```
LukasKalbertodt, line 134: `lol` should get a more descriptive name. Or, just rewrite it to now require a temporary variable.

```tsx
entries[value.name] = Object.entries(sequences).map(...);
```

src/main/SubtitleListEditor.tsx
```diff
@@ -276,23 +276,30 @@ const SubtitleListSegment = React.memo((props: subtitleListSegmentProps) => {
...
-      dispatch(setFocusSegmentTriggered(true))
-      dispatch(setFocusToSegmentBelowId({identifier: identifier, segmentId: cue.idInternal}))
-    },
-    delete: () => {
-      dispatch(setFocusSegmentTriggered(true))
-      dispatch(setFocusToSegmentAboveId({identifier: identifier, segmentId: cue.idInternal}))
-      deleteCue()
-    },
-  }
+  const hotkeyRef = useHotkeys<HTMLDivElement>([KEYMAP.subtitleList.addAbove.key, KEYMAP.subtitleList.addBelow.key, KEYMAP.subtitleList.jumpAbove.key, KEYMAP.subtitleList.jumpBelow.key, KEYMAP.subtitleList.delete.key],
```
LukasKalbertodt, line 279: This line should probably be broken into shorter ones.

src/main/SubtitleListEditor.tsx
```diff
@@ -276,23 +276,30 @@ const SubtitleListSegment = React.memo((props: subtitleListSegmentProps) => {
...
+          dispatch(setFocusSegmentTriggered(true))
+          dispatch(setFocusToSegmentBelowId({identifier: identifier, segmentId: cue.idInternal}))
+          break;
+        case KEYMAP.subtitleList.delete.key.split('+').pop():
+          dispatch(setFocusSegmentTriggered(true))
+          dispatch(setFocusToSegmentAboveId({identifier: identifier, segmentId: cue.idInternal}))
+          deleteCue()
+          break;
+      }
+    }, { enableOnFormTags: ['input', 'select', 'textarea'] }, [identifier, cue, props.index])
```
LukasKalbertodt, line 302: Can't this be multiple `useHotkeys` calls? Here you are calling it once, with 5 different buttons and in the handler you `switch` over the actual key pressed. Unless I'm missing something having 5 `useHotkeys` calls is way easier, no?

src/main/SubtitleListEditor.tsx
```diff
@@ -367,84 +374,82 @@ const SubtitleListSegment = React.memo((props: subtitleListSegmentProps) => {
...
-          onKeyDown={(event: React.KeyboardEvent) => {
-            if (event.key === "Enter" && !event.shiftKey) {
-              event.preventDefault()
-              addCueBelow()
-            }
-          }}
-          onChange={updateCueText}
-          onFocus={() => setTimeToSegmentStart()}
-        />
+    <div ref={hotkeyRef} css={segmentStyle} tabIndex={-1} style={{
```
LukasKalbertodt, line 377: Uhm am I see this correctly that you are using `style` which is directly translated to the `style` HTML attribute? Shouldn't this use `css`? You can simply interpolate the `segmentStyle` into that object.

src/main/SubtitleTimeline.tsx
```diff
@@ -73,12 +73,10 @@ const SubtitleTimeline: React.FC = () => {
...
-  const handlers = {
-    left: () => dispatch(setCurrentlyAt(Math.max(currentlyAt - keyboardJumpDelta, 0))),
-    right: () => dispatch(setCurrentlyAt(Math.min(currentlyAt + keyboardJumpDelta, duration))),
-    increase: () => setKeyboardJumpDelta(keyboardJumpDelta => Math.min(keyboardJumpDelta * 10, 1000000)),
-    decrease: () => setKeyboardJumpDelta(keyboardJumpDelta => Math.max(keyboardJumpDelta / 10, 1))
-  }
+  useHotkeys(KEYMAP.timeline.left.key, () => dispatch(setCurrentlyAt(Math.max(currentlyAt - keyboardJumpDelta, 0))), {}, [currentlyAt, keyboardJumpDelta]);
+  useHotkeys(KEYMAP.timeline.right.key, () => dispatch(setCurrentlyAt(Math.min(currentlyAt + keyboardJumpDelta, duration))), {}, [currentlyAt, keyboardJumpDelta, duration]);
+  useHotkeys(KEYMAP.timeline.increase.key, () => setKeyboardJumpDelta(keyboardJumpDelta => Math.min(keyboardJumpDelta * 10, 1000000)), {}, [keyboardJumpDelta]);
+  useHotkeys(KEYMAP.timeline.decrease.key, () => setKeyboardJumpDelta(keyboardJumpDelta => Math.max(keyboardJumpDelta / 10, 1)), {}, [keyboardJumpDelta]);
```
LukasKalbertodt, line 79: I'd say those lines are also quite long and dense. Splitting them would increase readability IMO. This is true for multiple lines in this PR, which I won't comment on individually.

src/main/WorkflowSelection.tsx
```diff
@@ -76,7 +76,7 @@ const WorkflowSelection : React.FC = () => {
     return (
       <div css={workflowSelectionStyle}>
         <h2>{topTitle}</h2>
-        {topText}
+        {!!topText}
```
LukasKalbertodt, line 79: Hu, what's that for? You convert it to a boolean, but rendering booleans in JSX results in nothing.

src/main/KeyboardControls.tsx
```diff
@@ -95,20 +88,18 @@ const Entry: React.FC<{params: KeyMapDisplayOptions}> = ({params}) => {
...
-                {length - 1 !== index ? <div css={orStyle}>+</div> : ''}
-              </>
-            ))}
-            <div css={orStyle}><Trans>{arr.length - 1 !== index && t("keyboardControls.sequenceSeparator")}</Trans></div>
-          </div>
-        ))}
-      </div>
+      <div css={labelStyle}><Trans>{name || t("keyboardControls.missingLabel")}</Trans></div>
+      {sequences.map((sequence, index, arr) => (
+        <div css={sequenceStyle} key={index}>
```
LukasKalbertodt, line 93: You swapped the order of some stuff so that one `<div css={sequencesStyle}>` is created per sequence. This means each seqeuence is in its own line, which is different from before. It becomes clear what I mean when you compare the shortcut overview pages of the current main branch and this.

LukasKalbertodt (changes requested):
Finally got around to reviewing this. Many of those inline comments are just code-style things that you are free to decide what to do with :P 

From testing the deployment I noticed a few more things:

- The order of boxes in the shortcut overview changed. Not a big problem, just making sure you are aware.
- ctrl alt d shows the desktop on Ubuntu :P Again, not a fault of this PR, just FYI
- ctrl alt c does not work for me at all, but neither does `editor.opencast.org`. Tested in Chrome and Firefox.
  - In fact, it seems like no ctrl+alt shortcut works for me. Not the jumping around ones either. Not for the current main deployment either. Uhhhm, is that just my desktop environment?

Studio uses `react-hotkeys-hook` too btw, that's good.

src/globalKeys.ts
```diff
@@ -25,159 +23,108 @@ const rewriteKeys = (key: string) => {
   if (isMacOs) {
     newKey = newKey.replace("Alt", "Option")
   }
+  newKey = isMacOs ? newKey.replace("Mod", "Command") : newKey.replace("Mod", "Control")
```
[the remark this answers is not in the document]
Arnei: Yup, this is purely for display purposes.

src/index.tsx
```diff
@@ -26,24 +31,20 @@ const initialize = Promise.race([
 ]);
 
 const render = (body: JSX.Element) => {
-  ReactDOM.render(body, document.getElementById('root'));
+  root.render(body);
 };
```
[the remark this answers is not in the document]
Arnei: Used below in line 52. Admittedly not the cleanest code.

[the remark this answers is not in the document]
LukasKalbertodt: Ah wupps, yeah indeed. But yeah so either I would use this `render` function in both cases (the ok and error case) or just remove it and replace the `render(...)` call below with `root.render(...)`

src/main/SubtitleListEditor.tsx
Arnei: Probably. I don't know enough about `ref`s to properly assign multiple of them to a single `div`, so this was simply easier to write for me ^^'

src/main/WorkflowSelection.tsx
Arnei: I don't even know. Will undo.

Arnei:
> The order of boxes in the shortcut overview changed. Not a big problem, just making sure you are aware.

Order was arbitrary before, is still arbitrary now. Thanks for the notice tho :D

> ctrl alt d shows the desktop on Ubuntu :P Again, not a fault of this PR, just FYI

Should really get to allowing users to set their own shortcuts. Is this something we want to solve in appkit?

> In fact, it seems like no ctrl+alt shortcut works for me. Not the jumping around ones either. Not for the current main deployment either. Uhhhm, is that just my desktop environment?

#works-for-me. We had users reporting shortcuts not working in the past, but I could never replicate their issues (with browserstack). If you manage to ever pinpoint what's causing issues for you it might help those other users as well.

src/main/KeyboardControls.tsx
Arnei: That is intended. I want to render each sequence in its own line to help with readability. This was discussed in some other issue, and I simply took the opportunity since I was working on that code anyway. But admittedly, this has nothing to do with updating to React 18, so I can try to revert those changes and make a separate PR for them, should you deem it desirable.

LukasKalbertodt: Ah I understand. Yeah no problem, keep it as part of this PR, I don't mind. Maybe the PR description should mention that though and maybe link to said discussion.

LukasKalbertodt (approved):
All code stuff was resolved. And now the new shortcuts also work for me. With the exception of shift+alt+i which opens a Chrome report window in Chrome.

PR #1060, opened 2023-05-22
Add volume control

### LINKED ISSUE #80: Add volume control
It may be nice to be able to adjust the volume of the video, instead of having to rely on global or browser volume controls.

dennis531:
Add a slider below the video to adjust the volume:

image

Closes #80


Arnei (changes requested):
Thanks for taking this on.

It irks me that the Play Button is not centered anymore. Have you tried other layouts?

This might be one of the few cases where we can use an icon with no text, instead of only text. The audio control icon is pretty universal.

Dark mode stuff:
Bildschirmfoto vom 2023-05-23 16-44-37
Not sure why the handle is transparent here. The blue highlighting glow is also barely visible. Also, this makes for a stylistic inconsistency, as no other element in the ui glows blue when hovered. Thoughts about making this more in-line?

The audio level is persisted between different views which I like, but the slider does not reflect that. It will always reset to 100% upon switching views.

dennis531:
What do you think about this layout?
image
image

Arnei:
Icons are good. Maybe one would be enough? No hard opinion on that. Could make the left icon clickable for easy mute/unmute, but that's more of an extra and not required for this PR to go through.

Persisting volume between views works.

Visual feedback for hovering/focusing on the slider is now barely visible in light mode and completely illegible in dark mode.

Still not fond of the arrangement of the controls. But I also have no idea how to keep the playbutton centered and order the other controls in an intuitive fashion. Would be nice if some design-inclined person could take a look at this.

dennis531:
The volume icon can be used to mute the video. 

I hope the visual feedback for hovering on the slider is now more expressive.

I hope that we will receive feedback about the arrangement of the controls.

ziegenberg:
Hi @dennis531, could you rebase your PR?

dennis531:
A rebase wasn't possible as some files were no longer present. Therefore, I re-implemented the changes based on the diff and adopted the appearance to the new design.

Arnei (approved):
Thanks for doing this for the redesign. Personally, I'm happy with how it looks and works. Will wait with merging a couple days so others can check it out too.

ziegenberg:
The two bars with the different controls are not aligned, which looks odd. This comes from trying to centre the play button. Which now is no longer centred.

Before:
Screenshot from 2023-09-27 15-47-06

Now:
Screenshot from 2023-09-27 15-46-37

Play button centred:
Screenshot from 2023-09-27 15-46-32


I propose to just centre the two bars with the controls and do not try to centre the playbutton.

dennis531:
I also prefer the solution with the centered bar.

Centered Bar:
image

ziegenberg:
Looks good to me!

ziegenberg (approved):

PR #1077, opened 2023-06-08
Fix some eslint warnings to get closer to other projects (Studio, Tobira)

LukasKalbertodt:
This is preparation for our upcoming plans to publish one shared ESLint config that can be used by many React+TS Opencast projects. But certainly for Editor, Studio and Tobira. I will explain more about this whole plan and why we think it's a good idea in the upcoming technical meeting on Tuesday. On that meeting I also want to ask the community if we can publish said configuration under the `@opencast` scope on NPM.

But before these things have been decided, I already ran eslint on this project with said config. In this PR I fixed many of the warnings that I think were just oversights/typos and that I assume you would agree should be fixed. If that is not the case, please let me know. I can drop individual commits from this PR.

Due to the pending decisions mentioned above, this PR does not yet refer to this new eslint config. I just did this locally. The PR just contains fixes.

Reviewing this is always weird, but it's probably best to go commit by commit. Maybe? At least the commit messages contain some useful information.


Arnei (approved):
All changes look reasonable and like they would improve readability in general.

This is gonna cause all the merge conflicts though....

LukasKalbertodt:
The last commit is not ready yet: it still contains a path dependency. That's also why CI fails. I will replace that with a real dependency as soon as it is decided where it should live. But yes, therefore marking this as draft.

Arnei (approved):
Approved again as it works and I agree with the changes made hence. We should work on some of those overrides in the future though.

PR #1096, opened 2023-07-03
Convert text to path in logo

lkiesow:
This patch converts the text included in the logo to a path. This will ensure that the logo is displayed correctly if the users does not have the font used for the text installed.


public/opencast-editor.svg
```diff
@@ -71,17 +73,28 @@
...
+     id="path182" /><path
+     d="m 524.8124,77.910515 c 8.55,0 15,-6.45 15,-15.0375 0,-8.55 -6.45,-15 -15,-15 -8.5875,0 -15.0375,6.45 -15.0375,15 0,8.5875 6.45,15.0375 15.0375,15.0375 z m 0,-2.25 c -7.275,0 -12.7875,-5.5125 -12.7875,-12.7875 0,-7.2375 5.5125,-12.75 12.7875,-12.75 7.2375,0 12.75,5.5125 12.75,12.75 0,7.275 -5.5125,12.7875 -12.75,12.7875 z"
+     style="fill:#ffffff"
+     id="path184" /><path
+     d="m 563.62489,75.810515 -6.45,-10.3125 c 3.7875,-1.0875 6.3,-4.2375 6.3,-8.3625 0,-5.2125 -3.9,-8.8875 -9.375,-8.8875 h -7.1625 c -0.675,0 -1.125,0.45 -1.125,1.125 v 27.0375 c 0,0.675 0.45,1.125 1.125,1.125 0.675,0 1.125,-0.45 1.125,-1.125 v -10.5 h 6.0375 0.675 l 6.9,11.025 c 0.225,0.3375 0.45,0.6 1.0125,0.6 1.1625,0 1.575,-0.7125 0.9375,-1.725 z m -15.5625,-12.15 v -13.1625 h 6.0375 c 4.1625,0 7.125,2.7375 7.125,6.6375 0,3.825 -2.9625,6.525 -7.125,6.525 z"
+     style="fill:#ffffff"
+     id="path186" /></g>
 <text
    xml:space="preserve"
    style="font-size:37.5px;line-height:1.25;font-family:'DejaVu Sans Mono';-inkscape-font-specification:'DejaVu Sans Mono';letter-spacing:0px;word-spacing:0px;stroke-width:0.9375"
```
LukasKalbertodt, line 100: Starting here, there is some weird out-of-bounds text box with the text "q". That should be removed as well.

LukasKalbertodt (changes requested):
Thanks! Can you also use the "export optimized SVG" option in Inkscape? That reduces the size of the logo. I also left this inline comment. 

I actually already created a fixed logo file on Thursday, but forgot to upload it publicly :see_no_evil: If you want you can just use that and update the PR with this:

opencast-editor-minified

public/opencast-editor.svg
lkiesow: Since you started complaining about unnecessary stuff in the SVG, I removed this and a buch of additional data Inkscape threw in there we don't need.

You can probably remove more if you want, but I think it's fine like this.

Arnei (approved):

Issue #1083, opened 2023-06-26
Add "remove all segments" function


kweingaertne:
In the old editor, there was a button to remove all the segments automatically made in a given recording. Having this back would be a great quality of life improvement. Also, it would be great to be able to restore or delete segments without moving the cursor into the segment first. In the old editor, there was a little icon on the top right corner of every segment, letting you delete/ restore them without interrupting playback.

Arnei:
Regarding the second part of this issue: I personally dislike the icons in the top right segment corner. They're small and thus hard to hit, they disappear if segments are too small, and are often just as far away from the cursor as the general delete button.  I think keyboard shortcuts are a better solution.

But you mentioned that would like to be able to delete/restore segments without interrupting playback. I had not considered this previously, could you describe your use case in greater detail? I'd be interested in hearing if others have the same use case.

dagraf:
I agree here with @Arnei: I like that we do not have the icons in the top right segment corner anymore. Deleting/restoring segments without interrupting playback is still possible if the option "Preview mode" is DEactivated.

ziegenberg:
If you click "Merge All" Segments, this removes all segments. Which is basically the description of this issue. I would consider this as "solved".

dagraf:
@ziegenberg: I guess your referring tho https://github.com/opencast/opencast-editor/pull/1097? If so, I would consider it as solved, too.

Arnei:
Cool, closing this then. If people want to further discuss the second part of this issue, they may open another issue for that.

Issue #943, opened 2022-12-24
Add right click menu for cutting


lkiesow:
Similar to the right-click context menu Element offers (see below), we should implement a context menu for the editor view to quickly remove or modify segments. This should be an additional menu though.

![Screenshot from 2022-12-24 10-54-57](https://user-images.githubusercontent.com/1008395/209430864-d604ae74-2c83-44a1-8c50-c5b83e833bcd.png)

PR #1097, opened 2023-07-04
Remove all segments button

dennis531:
Contains the following changes:
- Adds a button to delete all segments
- ~~Adds a button for each segment to delete corresponding segment~~

Ref #1083


lkiesow:
I think you misunderstood #1083. I think what @kweingaertne wants is not to mark the content for removal, but to be able to quickly remove all segments itself. Right now, you would do that by starting somewhere and hitting the “Merge Left/Right” buttons over and over again.

Marking all segments for removal doesn't really make sense because you would end up with a 0 seconds long video file which should then fail to process as Opencast would tell you that this doesn't really make sense :D

lkiesow:
I'm also not the biggest fan of the icon on the timeline. The old editor had several of them and I would like to not see them creep back one after another. Maybe instead as a “pro” feature to easily access the functionality implement #943 instead?

What do you think, @kweingaertne?

lkiesow:
Finally (for the next time), having one feature/change pre pull request makes these changes easier to discuss, review and merge.

dennis531:
Now the button allows to merge all segments into a single segment. 
Additionally, the delete button on each segment is removed.

dennis531:
> Usually when merging segments, the resulting segment inherits the "deleted state" of the previously selected segment. So if I select a "blue" segment and merge it with "red", the resulting segment is "blue". This is not true for the specific case in which I select a "red" segment" and merge it _left_ with a "blue" segment. The resulting segment is "blue" instead of "red". Could you fix that?

Thanks for noticing. It should be fixed now.

Arnei (approved):

dennis531:
> > Usually when merging segments, the resulting segment inherits the "deleted state" of the previously selected segment. So if I select a "blue" segment and merge it with "red", the resulting segment is "blue". This is not true for the specific case in which I select a "red" segment" and merge it _left_ with a "blue" segment. The resulting segment is "blue" instead of "red". Could you fix that?
> 
> Thanks for noticing. It should be fixed now.

During solving the merge conflicts, I have noticed that this problem was still present. Hopefully, this is solved now.

Arnei (approved):

Issue #1084, opened 2023-06-29
Provide return link for integration


lkiesow:
Similar to Studio, it would be nice if a system integrating the Editor could provide a return link to the editor via URL parameter which would then be shown to users when they are finished editing and when they would currently be stuck in a dead end (e.g. after starting the processing):

![27d77d15-215b-44fc-960e-66e008758746](https://github.com/opencast/opencast-editor/assets/1008395/2411e8e1-56e5-4ab1-ac0b-34c8011bb9a3)

Which systems are allowed should be a configuration option in the editor configuration file so that you cannot generate an editor link which at the end will lead you to a malicious site. But this can be a very simple list of URL prefixes like `["https://opencast.org/", "https://university.edu/"]`.

dennis531:
I am not really sure whether the callback url should be an option in the config file. I think this should not be a problem, since URL parameters have a higher priority and the option in the file could be seen as a default value.

On the other hand, the list of allowed domains/prefixes should be configurable only in the file. Otherwise, this mechanism would not provide any protection.

dennis531:
Preview:
![image](https://github.com/opencast/opencast-editor/assets/44410838/c5b27e9b-4404-40d9-8412-9108ff894302)

snoesberger:
This is also an important feature for our institution, as users sometimes need to edit several videos in a row and want to go back to where they started. Or they may simply want to pick up where they left off before they started the editor.


> I think `Back to LMS` is too specific as we don't know if the return link refers to a LMS. I would suggest to name it more generic, e.g. `Back to previous system`.

I agree with that. It would be even better to make the button name configurable. You could specify a name in the config file for each permitted URL or prefix. This name then would be displayed as button text:

"Tobira": "https://tobira.opencast.org" --> Button text = "Back to Tobira"
"ILIAS": "https://ilias.university.edu" --> "Back to ILIAS"

ziegenberg:
@Arnei, I think this can be closed since #1100 was merged.

PR #1100, opened 2023-07-05
Add callback url for integration (#1084)

dennis531:
After finishing the editing, the user can now navigate back to the calling system using a button. For this, the integrating system can pass a `callbackUrl` url parameter to provide a return link. In the editor settings, the prefixes of the allowed callback urls must be provided in the `allowedCallbackPrefixes` option to prevent malicious callback urls.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -43,6 +43,7 @@ interface iSettings {
   id: string | undefined,
   allowedCallbackPrefixes: string[],
   callbackUrl: string | undefined,
+  callbackSystem: string | undefined,
   opencast: {
     url: string,
     name: string | undefined,
@@ -80,6 +81,7 @@ const defaultSettings: iSettings = {
   id: undefined,
   allowedCallbackPrefixes: [],
   callbackUrl: undefined,
+  callbackSystem: undefined,
   opencast: {
     url: window.location.origin,
     name: undefined,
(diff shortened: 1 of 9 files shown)


src/config.ts
```diff
@@ -146,6 +150,11 @@ export const init = async () => {
   // Prepare local setting to avoid complicated checks later
   settings.opencast.local = settings.opencast.local && settings.opencast.url === window.location.origin;
 
+  // Prevent malicious callback urls
+  settings.callbackUrl = settings.allowedCallbackPrefixes.some(
```
lkiesow, line 154: Here it is called `allowedCallbackPrefixes` in the configuration file and earlier in the code, it is `allowedCallbackDomains`. Since your example contains the URL schema, I think that prefix makes more sense.

But in any case, I think this will not work like this. You have to use one or the other.

Not that I've tested this :D

dennis531: Thanks for pointing it out. Originally, I called this setting `allowedCallbackDomains` and later I renamed it to `allowedCallbackPrefixes`. Unfortunately, I missed this line.

dennis531: Now this option should be named persistently `allowedCallbackPrefixes`.

Arnei (approved):

dennis531:
Futher changes:
- Adds `callbackSystem` option to display the name of the callback system in the callback button.

Arnei (approved):

PR #1101, opened 2023-07-06
fixes #36, #997 respect workflow displayOrder, add filter workflows

### LINKED ISSUE #36: restrict workflows available in the editor

JamesUoM:
Respect workflow displayOrder
Add filter workflows by displayOrder  (min/max config settings)

UoM: MAT-477, MAT-478

### DIFF
--- a/editor-settings.toml
+++ b/editor-settings.toml
@@ -165,3 +165,12 @@
 # Type: boolean
 # Default: true
 #local =
+
+[workflow]
+
+# Filter the publishing workflows by their displayOrder value
+# min <= displayOrder <= max
+# Type: number
+# Default: undefined (unbound)
+#min =
+#max =
--- a/src/config.ts
+++ b/src/config.ts
@@ -57,6 +57,10 @@ interface iSettings {
     languages: { [key: string]: string } | undefined,
     icons: { [key: string]: string } | undefined,
     defaultVideoFlavor: Flavor | undefined,
+  },
+  workflow: {
+    min: number | undefined,
+    max: number | undefined,
   }
 }
 
@@ -92,6 +96,10 @@ const defaultSettings: iSettings = {
     languages: {},
     icons: undefined,
     defaultVideoFlavor: undefined,
+  },
+  workflow: {
+    min: undefined,
+    max: undefined
   }
 }
 let configFileSettings: iSettings
@@ -287,6 +295,11 @@ const types = {
       throw new Error("is not a string, but should be");
     }
   },
+  'number': (v: any, _allowParse: any) => {
+    if (isNaN(v)) {
+      throw new Error('is not a number, but should be');
+    }
+  },
   'boolean': (v: string, allowParse: any) => {
     if (typeof v === 'boolean') {
       return;
@@ -373,6 +386,10 @@ const SCHEMA = {
   thumbnail: {
     show: types.boolean,
     simpleMode: types.boolean,
+  },
+  workflow: {
+    min: types.number,
+    max: types.number
   }
 }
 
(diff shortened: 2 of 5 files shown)


JamesUoM:
deploy.opencast.org only has one workflow. To help with testing you can inject a dummy workflow by editing 
`src/main/WorkflowSelection.tsx` 
``` 
    const filterWorkflows = (workflowFilter: IWorkflowConfiguration | undefined, workflows: Workflow[]) => {
    // DEBUG
    let w : Workflow = {
      id: "1235",
      name: "test workflow",
      description: "bobbins",
      displayOrder: 600
    }

    workflows = [workflows[0], w]
    ...

**Review by Arnei — CHANGES_REQUESTED** [review]
Could you add the new settings to the `public/editor-settings.toml` as well?

I can crash the editor with the following setting 
```
min = 1001
max = 1000
```
Results in this error:
```
Uncaught TypeError: Cannot assign to read only property '0' of object '[object Array]'
    at Array.sort (<anonymous>)
    at WorkflowSelection (WorkflowSelection.tsx:58)
```
I have four workflow configured, with displayOrder values `1010`, `1000`, `100` and `no value`.

**Comment by Arnei** [comment]
Small remark: The editor has been respecting displayOrder, but apparently been sorting the wrong way around: https://github.com/opencast/opencast-editor/blob/897d4f5c59052eb00fefc857ed554802313ec3f9/src/redux/videoSlice.ts#L215. I suppose that code can go then.

File: src/config.ts [hunk 1]
```diff
@@ -287,6 +295,11 @@ const types = {
       throw new Error("is not a string, but should be");
     }
   },
+  'number': (v: any, _allowParse: any) => {
+    if (isNaN(v)) {
+      throw new Error('is not a number, but should be');
```
Arnei, line 300: Does not seem to thrown if I set `min` or `max` to a string. Should it?

JamesUoM:
> I have four workflow configured, with displayOrder values `1010`, `1000`, `100` and `no value`.

It's not the min>max but something weird triggered by the sort() when a displayOrder is undefined. You can't sort "in place" and need to create a copy. 

min>max results in zero workflows selected and so the warning is issued and all workflows are returned.

src/config.ts
Arnei: Welp, you're right. Don't know how I missed that.

editor-settings.toml
```diff
@@ -165,3 +165,12 @@
...
 #local =
+
+[workflow]
+
+# Filter the publishing workflows by their displayOrder value
+# min <= displayOrder <= max
+# Type: number
+# Default: undefined (unbound)
+#min =
+#max =
```
lkiesow, line 176: I'm not sure if this really makes sense and would like to hear your reasoning behind this. I find this extremely unintuitive. Why should the order determine if a workflow is being shown or not? As an admin, I could see it easily happen that I want to change the order and this causes the workflow to accidentally disappear.

More than that, with the old editor being sunsetted and about to be removen, the new editor is the only place in Opencast where the tag `editor` is being used. This means that you can just tag the workflows and only the workflows you want to be shown as `editor`.

This is much more straightforward ans also means that the same functionality is not being duplicated (you are essentially trying to add a second configuration option which does the same thing as an existing one).

That means, unless you have a good argument for adding this, I really think we shouldn't add this.

lkiesow (changes requested):
No complaints about respecting display order, but I'm really not sure about the filtering.

JamesUoM:
Fair point, this may not be for everyone.  We would have to address the issue of unwanted editor workflows (ie the default ones) in the end-user editor by other means.

--
------------------------------------------------------------------------
  James S. Perrin

  Media Technologies Team
  Office One, Kilburn Building,
  The University of Manchester
  Oxford Road, Manchester, M13 9PL

  e: ***@***.***
________________________________

KatrinIhler:
@JamesUoM I agree with Lars here, I don't think the filtering is something we want in the upstream. But the rest of the changes seem useful, could you maybe update this PR or file those separately?

JamesUoM:
To keep it clean and tidy I've created a replacement PR https://github.com/opencast/opencast-editor/pull/1179

PR #1102, opened 2023-07-12
Implement editor locking so only one person can edit a recording a time

JamesUoM:
This works with the corresponding feature in Opencast [1] when it is enabled in the editor service config. When locking is enabled only one person can edit a video at time. This is for the use case where multiple people have access to a video and may have been notified it is ready for editing.

[1] https://github.com/opencast/opencast/pull/4658

UoM: MAT-468


Arnei (approved):
Code looks good and works in testing. Gets an "Approved" from me!

KatrinIhler:
@JamesUoM Could you resolve the conflicts so we can merge this?

PR #1158, opened 2023-09-14
Align track view button placement with thumbnail view

### LINKED ISSUE #1157: Button Placement in Track and Thumbnail
In the track selection view, the button is below the preview image.
In the thumbnail view, the buttons are to the right of the generated image (unless the screen width becomes relatively small)

grafik
vs.
grafik

This feels inconsistent. Olaf suggested in #1079 to uniformly have the buttons to the right, with a vertical alignment like in the thumbnail view.

Arnei:
Makes the button in the track selection view behave more similar to how the thumbnail generation buttons behave.

Since there's only one button ("Delete Track"),
I'm a little worried if it still parses as a button without a border.

Aims to resolve #1157.


owi92 (approved):
Looks good to me.
One thing I noticed when looking at these buttons is that their icon color isn't inverted on hover in high contrast modes. But that's neither introduced by nor subject of this PR - I'll open a separate issue.

PR #1160, opened 2023-09-26
More small screen responsiveness

### LINKED ISSUE #23: Non Breaking Cutting Tools
The line of cutting tools is non-breaking, causing the whole app to effectively have a minimum width. 
In case of smaller screens, we could

- Hide less important buttons in a …-menu
- Reduce the buttons to their icons

Arnei:
Makes the editors appearance on narrow screens a little more palatable. This PR is not intended to make the editor fully usable in portrait mode, but rather help users with narrow screens and/or high zoom levels to get a little more out of their screen space.

Only touches the header, main menu, cutting view and metadata view for now.

Resolves #23, maybe?


src/main/Header.tsx
```diff
@@ -112,6 +124,7 @@ const Logo: React.FC = () => {
 
   const { t } = useTranslation();
   const { scheme } = useColorScheme();
+  const { width } = useWindowDimensions();
```
geichelberger, line 127: Do we need this? Else this line and the useWindowDimensions() import can be removed.

geichelberger (changes requested):
LGTM, except for the not-used value.

src/main/Header.tsx
Arnei: Removed it, thanks for catching that.

geichelberger (approved):
LGTM

PR #1183, opened 2023-10-20
Fix subtitles not being saved to Opencast

### LINKED ISSUE #1161: Subtitles get lost when clicking "Start processing"
Affects editor shipped with Opencast 14.4.

If an user adds subtitles to the video and clicks "Start processing" afterwards, the subtitles are not saved - in contrast to eventually added cut marks, that are stored and processed properly. (In fact a smil file is always created and saved, also when I did not added cutmarks).
When refreshing the editor, the subtitles are gone.

If "Save changes" is clicked, the subtitles are saved. When refreshing the editor, I can see the subtitles and clicking "Start processing" works: subtitles are picked up by both of my testing workflows (publish-assets and cut-and-publish) and they're successfully published to the player.

Arnei:
Fixes #1161.

Turns out the post request for starting processing in Opencast simply did not include the (changes to) subtitles. Now it does.

### DIFF
--- a/src/redux/workflowPostAndProcessSlice.ts
+++ b/src/redux/workflowPostAndProcessSlice.ts
@@ -17,7 +17,12 @@ export const postVideoInformationWithWorkflow = createAsyncThunk('video/postVide
   }
 
   const response = await client.post(`${settings.opencast.url}/editor/${settings.id}/edit.json`,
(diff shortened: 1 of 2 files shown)


KatrinIhler (approved):
Does what it says on the tin.

PR #1208, opened 2023-11-07
Don't show certain thumbnail buttons for single track

### LINKED ISSUE #1206: Skip the "Use for all tracks" option in "Thumbnails" if there is only one track
grafik

Arnei:
The "Generate All" and "Use for all tracks" buttons in the thumbnail view don't really serve a purpose when there is only one track. "Generate All" does the same thing as "Generate", and "Use for all tracks" does straight up nothing. Therefore, we might as well not display them if there is but a single track.

Fixes  #1206.

### DIFF
--- a/src/main/Thumbnail.tsx
+++ b/src/main/Thumbnail.tsx
@@ -174,7 +174,9 @@ const ThumbnailTable : React.FC<{
       )
     } else {
       return (<>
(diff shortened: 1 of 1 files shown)


LukasKalbertodt (approved):
Code looks good and it seems to work. (See https://test.editor.opencast.org/1208/2023-11-07_09-13-50/?mediaPackageId=ID-nasa-earth-4k)

PR #1209, opened 2023-11-07
Don't show track view if there is only one track

### LINKED ISSUE #1205: Skip the "Tracks" section of the editor if there is only one track
Cf. https://oc.tobira.ethz.ch/editor-ui/index.html?mediaPackageId=25e82f02-db10-4ba6-937f-3252353cfbe8. 

grafik

Or is there a reason to offer no option there?

Arnei:
If an event has only a single track, you cannot deselect that track, so the track view is rather pointless in that case. Therefore, we might as well not show it.

Fixes #1205.


LukasKalbertodt (approved):
Code looks good and seems to work (see https://test.editor.opencast.org/1209/2023-11-07_09-24-17/?mediaPackageId=ID-nasa-earth-4k)

lkiesow:
It might still be confusing to users when part of the tooling just silently disappears.
Maybe just adding an info box to the track selection would actually be better?
There you could put that this tool is disabled and why.

KatrinIhler (changes requested):
Here only the info box thingy is still outstanding. Requesting changes to accurately represent the state.

Arnei:
Basic info box. Suggestions for phrasing are welcome :)

Bildschirmfoto vom 2024-05-13 15-45-26

src/i18n/locales/en-US.json
```diff
@@ -240,6 +240,14 @@
...
     "trackSelection": {
       "title": "Select track(s) for processing",
+      "trackInactive": "inactive",
+      "deleteTrackText": "Delete Track",
+      "restoreTrackText": "Restore Track",
+      "cannotDeleteTrackText": "Cannot Delete Track",
+      "deleteTrackTooltip": "Do not encode and publish this track.",
+      "restoreTrackTooltip": "Encode and publish this track.",
+      "cannotDeleteTrackTooltip": "Cannot remove this track from publication.",
+      "noSelectionPossible": "Track selection is disabled for this video, because there are not enough tracks to deselect any.",
```
dennis531, line 250: I would suggest providing a more precise description about the number of tracks required:

```suggestion
      "noSelectionPossible": "Track selection is disabled for this video, because at least two tracks are required to deselect any.",
```

KatrinIhler:
Repointed to develop because the new track selection was removed from 17 by #1589.

KatrinIhler (approved):

Issue #1207, opened 2023-11-06
Adapt the size of the middle canvass in "Thumbnails"


oas777:
![grafik](https://github.com/opencast/opencast-editor/assets/1752342/bd512ed1-0b18-4042-8fa0-3b64c2c5a7dc)

Arnei:
Could you explain in greater detail what your issue is?

oas777:
You have three canvasses and one of them is not as wide as the others. Why is that? Because my argument (with different tickets also) is coherence, coherence, coherence.

Arnei:
So if I get that correctly you want the box/outline around the video in the middle to have the same width as the box/outline above and below it, okay.

oas777:
Yep, thanks.

Arnei:
The thumbnail view has been overhauled, so this is now outdated I believe.

PR #1210, opened 2023-11-07
Thumbnail make boxes same size

Arnei:
Currently in the thumbnail view, the videos are the same size as the thumbnails. However, the outlines surrounding them are different, as the outline around the thumbnails also contains buttons. This enlarges the videos to make the outlines have the same width.

Partially addresses #1207. From a usability point-of-view, I don't think making the timeline the same width as the stuff above it is currently reasonable. Since we cannot zoom into the timeline (yet™), it should be as wide as possible.

### DIFF
--- a/src/main/Thumbnail.tsx
+++ b/src/main/Thumbnail.tsx
@@ -106,7 +106,8 @@ const Thumbnail : React.FC = () => {
         discard={discardThumbnail}
       />
       <div css={bottomStyle}>
-        <VideoPlayers refs={generateRefs} widthInPercent={100}/>
+        {/* use maxHeightInPixel to make video players the same size*/}
+        <VideoPlayers refs={generateRefs} widthInPercent={100} maxHeightInPixel={376}/>
         <div css={videosStyle(theme)}>
           <Timeline
             timelineHeight={125}
--- a/src/main/Thumbnail.tsx
+++ b/src/main/Thumbnail.tsx
@@ -101,6 +101,7 @@ const Thumbnail : React.FC = () => {
       <div css={[titleStyle(theme), titleStyleBold(theme)]}>{t('thumbnail.title')}</div>
       <ThumbnailTable
         inputRefs={inputRefs}
+        generateRefs={generateRefs}
         generate={generate}
         upload={upload}
         uploadCallback={uploadCallback}
@@ -108,7 +109,7 @@ const Thumbnail : React.FC = () => {
       />
       <div css={bottomStyle}>
         {/* use maxHeightInPixel to make video players the same size*/}
-        <VideoPlayers refs={generateRefs} widthInPercent={100} maxHeightInPixel={376}/>
+        <VideoPlayers refs={generateRefs} widthInPercent={100} maxHeightInPixel={420}/>
         <div css={videosStyle(theme)}>
           <Timeline
             timelineHeight={125}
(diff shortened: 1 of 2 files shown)


src/main/VideoPlayers.tsx
```diff
@@ -37,7 +37,7 @@ const VideoPlayers: React.FC<{refs: any, widthInPercent?: number}> = ({refs, wid
     borderRadius: '5px',
     ...(flexGapReplacementStyle(10, false)),
 
-    maxHeight: '300px',
+    maxHeight: maxHeightInPixel + 'px',
```
LukasKalbertodt, line 40: Editor is also using emotion-js, right? You should be able to just pass integers here. Also the `borderRadius: 5` above should just work. But if the whole codebase uses `+ 'px'` already, then it's probably not necessary to fix this for this one PR now.

LukasKalbertodt (approved):
Seems to work! I agree that the timeline should stretch the full width.

src/main/Thumbnail.tsx
```diff
@@ -106,7 +107,8 @@ const Thumbnail : React.FC = () => {
         discard={discardThumbnail}
       />
       <div css={bottomStyle}>
-        <VideoPlayers refs={generateRefs} widthInPercent={100}/>
+        {/* use maxHeightInPixel to make video players the same size*/}
+        <VideoPlayers refs={generateRefs} widthInPercent={100} maxHeightInPixel={376}/>
```
LukasKalbertodt, line 111: This magic number depends on the width of the buttons. If a button is hidden (like in #1208), the width changes. 

image

So this should likely have a different solution. Maybe both boxes just get a fixed width. Or something else.

Arnei: Welp, that teaches me to not go for the quick and dirty solution. I'll see if I can get a proper solution working, but that'll require some time.

src/main/VideoPlayers.tsx
Arnei: It does use Emotion, but not emotion-js specifically I think? Anyway, I like being specific about my units :D

src/main/Thumbnail.tsx
```diff
@@ -209,15 +213,19 @@ const ThumbnailTableRow: React.FC<{
...
   uploadCallback: any,
   discard: any,
-}> = ({track, index, inputRefs, generate, upload, uploadCallback, discard}) => {
+}> = ({track, index, inputRefs, generateRef, generate, upload, uploadCallback, discard}) => {
 
   const { t } = useTranslation()
   const theme = useTheme();
 
+  // The "+40" comes from padding that is not included in the "getWidth" function
+  const videoWidth = generateRef ? generateRef.getWidth() + 40 : undefined
```
LukasKalbertodt, line 227: Would using `offsetWidth` instead of `clientWidth` remove the need to add 40 manually?

LukasKalbertodt (commented):
I am not sure this approach is the best solution. This gets the width once in the render code and is only updated on rerender. Maybe the width can change without a rerender happening? E.g. text wrapping, other CSS wrapping or screen-width based logic? I feel a fully CSS-based solution might be better, i.e. wrapping everything in a flex or grid and then let one child control the column width, the other child just filling the column. 

However, it seems to work and I couldn't find any problems in practice. And I know this kind of sizing via CSS can be very tricky to get right, I struggled a lot with it and even found browser bugs. So I don't want to block merging on this theoretical concern.

src/main/Thumbnail.tsx
Arnei: It did not

LukasKalbertodt: Sad :shrug:

PR #1230, opened 2023-12-14
Split track selection to video and audio selection

### LINKED ISSUE #1009: Remove video or audio from a track
The track view allows users to "delete" tracks if there are multiple tracks in the event. However, it would be nice if the track view would allow to remove only the video or only the audio from a track. For example, you might want to remove the video from a presenter track if the presenter does not want to appear in the recording, but keep the audio from the track.

Arnei:
In the "Tracks" view, instead of de-/selecting the whole track, the video and audio streams can be de-/selected individually. At least one video stream must remain selected. All audio streams can be deselected. If the audio stream is marked as unavailable, it cannot be de-/selected.

Shows the waveform from the timeline for a graphical representation of the audio stream. However, there will only be one waveform generated for the timeline, so in case of additional tracks a placeholder image will be shown instead. Any suggestions/submissions for a better placeholder graphic are very welcome.

Resolves #1009. #1009 also mentions potential issues in the backend, while this PR only addresses the frontend. From my limited testing the backend seems fine when using the default community workflows (no errors, the correct streams get published). If there are any users around that are still using the track selection feature in the old editor, and that are aware of any backend issues, please do tell me about them.


Arnei:
Possible way to go about the usability issues you highlighted:

1. Leave it as is, people will "somehow just know" how it works at their place. Arguably workable, but not a great solution.

2. Introduce restrictions on what combination of tracks can and cannot be selected. Would need to be configurable, because there will always be admins that want to allow combinations that others would restrict, like the ones you mentioned. But then we again run into the issue of having to "just know" what certain combinations will do.

3. Add more buttons/checkboxes/ui that describes what should happen with the tracks, e.g. `Mux all remaining audio streams onto A/B/C` or `Keep audio on tracks`, `Allow tracks with only audio`. Arguable the way to go even if it muddles up the UI somewhat, but also requires an editor workflow to support all those new flags.

lkiesow:
The original goal of the editor was to be simple since it is meant for non-technical end-users. I would really like to keep that goal since we expose the editor to such users.

Therefor, a lot of implicit knowledge as a requirement isn't a great solution and neither is having a lot of boxes and buttons with technical terms.

As an alternative, we could introduce one simple way to handle the result. My suggestion would be: If you remove *any* audio stream, *all* other audio streams will be mixed and attached to all videos. This rule is simple and works with as many videos tracks as you want. It also works well with other parts of Opencast like the Paella Player, which will just play back a single stream.

Running with that idea, we would need a change in the backend workflow operation and a hint about how this works in the UI of this pull request. The hint should also include that the mixing will happen if a track is already missing an audio stream.

Furthermore, since users should be able to deal with individual audio streams, we also need to make sure the editor can playback all streams, and individual streams. A simple solution may be playback controls on the tracks page where you then disable the audio streams of the tracks were audio has been removed?

KatrinIhler:
Finally getting around to commenting on this... I will not chime in with the ongoing discussion, but just say that 

> At least one video stream must remain selected

is something I would like to see either being configurable in the backend, or simply not enforced (which I think you already knew I was gonna say ;)), because it doesn't take into account audio-only use cases. Consider the following:

1. An institution with custom workflows for publication that can handle audio-only input (either by rendering a static image into it, or by having a player that can handle audio-only), and
2. A recording where the video is either black or otherwise damaged or unusable, but the audio is fine, and I want to deselect the video and publish the audio only with my custom workflow, or
3. A recording that doesn't come with a video stream in the first place that I want to cut and publish.

Both of these cases should be possible with the new editor. It doesn't have to be the case out of the box, but I shouldn't need to patch the code to make this work, just do some configuration. So basically the same arguments I made [here](https://github.com/opencast/opencast/pull/5023), except that for the new editor, making this configurable with an "allow audio only in editor" or something would actually be worth it.

luniki:
Lars send me this "oneliner" to create a triple event:

```bash
curl -u admin:opencast "${SERVER}/ingest/addMediaPackage/fast" \
        -F 'flavor=presenter/source' \
        -F mediaUri=https://example.com/opencast/test-media/goat.mp4 \
        -F 'flavor=presentation/source' \
        -F mediaUri=https://example.com/opencast/test-media/goat.mp4 \
        -F 'flavor=experiment/source' \
        -F mediaUri=https://example.com/opencast/test-media/goat.mp4 \
        -F identifier=three \
        -F title="Three Video Streams" \
        -F creator="$(shuf -n 1 /usr/share/dict/words)"

```

Arnei:
Results of todays discussion between Sascha, Katrin, Lars, Arne and Marcus:
- Should be able to select 0, 1 or more videos
- In the same vein, should be able to select 0, 1 or more audios
- At least one track (video or audio) needs to remain
- The same audio is applied to all videos:  None, one or multiple muxed
  - Have a note in the UI that informs the user about muxing  
- In case no video remains selected, the remaining audios are muxed into one audio track
- Default selection when opening up the tracks view: All videos are selected and only the first audio track is selected
  - ...why did we want this again?
- There should be a big confirm button
  - This idea here is to avoid a user making an accidental change to their event

Key-Image
Bildschirmfoto vom 2024-05-22 15-46-30

luniki:
I talked to @Arnei about this and created a first version of it. Here are some screenshots.

### Initial state after opening the "Tracks" tab

Screenshot 2024-05-29 at 12-04-11 Opencast Editor

### After de-selecting a video track

Screenshot 2024-05-29 at 12-08-49 Opencast Editor

### After de-selecting all video tracks

Screenshot 2024-05-29 at 12-11-03 Opencast Editor

### After de-selecting all but one track

Screenshot 2024-05-29 at 12-13-48 Opencast Editor

luniki:
The "Confirm selection" buttons seems to be useless. Whats the reasoning behind it?

snoesberger:
> The "Confirm selection" buttons seems to be useless. Last we talked we defined:
> 
> > There should be a big confirm button
> > * This idea here is to avoid a user making an accidental change to their event
> 
> The initial state is that all tracks are selected. This can also be easily determined in the code. So if we never de-select any tracks, we don't change the event?!

IIRC, the idea was not to have a confirm button, but to have a checkbox to activate the track selection. So if a user wants to de-select tracks, they first have to activate this "modify track selection" checkbox. As long as the checkbox is inactive, video and audio selection is grayed-out, nothing can be changed and the track selection will be completely ignored. As soon as the checkbox is activated, the user will be able to de-select video and/or audio tracks, and Opencast will process video and audio accordingly (muxing audio, etc.) as soon as the user starts the workflow. @KatrinIhler and @lkiesow please correct me if I'm wrong.

luniki:
I have pushed my changes to this PR to my own branch: https://github.com/luniki/opencast-editor/commits/tracks-select-audio/

luniki:
Here are some screenshot of my current state of track selection.

### Entering the track selection
image

### Enabled track selection
image

### De-selected all but one track
image

luniki:
The checkbox labelled "Customize track selection" is a new part of the state of the editor which has to be transferred to the Opencast EditorService endpoint. At the moment there is an additional boolean key in the JSON object.

This requires changes in the `EditorRestEndpointBase` class (and the GSON classes).

Or do I get this wrong? How should that checkbox be encoded?

luniki:
I pushed my changes to my forked branch of opencast-editor: https://github.com/luniki/opencast-editor/commits/tracks-select-audio/

In the backend I edited the `EditingData` class: https://gitlab.elan-ev.de/mlunzena/opencast-bern/-/commits/feature/track-selection

snoesberger:
Thanks for the update. I have tested the latest changes and have the following comments:

- If there is a track without audio, this track is always selected and can't be deselected. This track should be deselected by default, because otherwise it is possible to publish this silent audio track without video, which makes no sense.
grafik

- If the video on the left has no audio, a placeholder image will be displayed instead of the generated waveform for the video on the right:
grafik

- Rather than having "No Audio" (which is styled differently to the rest), I'd prefer to have a simple line indicating that there is no audio.
grafik

Arnei:
> If there is a track without audio, this track is always selected and can't be deselected. This track should be deselected by default, because otherwise it is possible to publish this silent audio track without video, which makes no sense.

To phrase it another way:
- For tracks without audio, the audio should initially be deselected
- "No Audio" can later be selected (it is not locked into being deselected)
- It should not be possible to create selections that only contain "No audio"

Does that seem right to you?

Arnei:
> Rather than having "No Audio" (which is styled differently to the rest), I'd prefer to have a simple line indicating that there is no audio.

Sure, will do. If you have any other suggestions for the "No Audio" or "Placeholder" images, feel free to post them. I did not put too much effort into them initially and have no attachment to how they look.

snoesberger:
> Sure, will do. If you have any other suggestions for the "No Audio" or "Placeholder" images, feel free to post them. I did not put too much effort into them initially and have no attachment to how they look.

Or why don't we just remove the checkbox and image completely? If there is no audio track we shouldn't show one. Instead we could just show a "rectangle" as a placeholder:
grafik

This would also make your questions from the comment https://github.com/opencast/opencast-editor/pull/1230#issuecomment-2154255479 obsolete. What do you think?

Arnei:
> If the video on the left has no audio, a placeholder image will be displayed instead of the generated waveform for the video on the right:

Might this be because there actually is no generated waveform? Currently, a waveform will only be generated for the first track, even if that track has no audio (#1376 should fix that), which would be presentation in your image.

Arnei:
>Or why don't we just remove the checkbox and image completely? If there is no audio track we shouldn't show one. Instead we could just show a "rectangle" as a placeholder:

Might be because of my developer glasses, but to me that looks like the page is broken :D

snoesberger:
> > If the video on the left has no audio, a placeholder image will be displayed instead of the generated waveform for the video on the right:
> 
> Might this be because there actually is no generated waveform? Currently, a waveform will only be generated for the first track, even if that track has no audio (#1376 should fix that), which would be presentation in your image.

But in my case there is already a waveform available, at least in the cutting section the waveform for the presenter is displayed.

Arnei:
> But in my case there is already a waveform available, at least in the cutting section the waveform for the presenter is displayed.

Okay, strange. I'll look into it, thanks for telling me.

snoesberger:
> > Or why don't we just remove the checkbox and image completely? If there is no audio track we shouldn't show one. Instead we could just show a "rectangle" as a placeholder:
> 
> Might be because of my developer glasses, but to me that looks like the page is broken :D

You may be right, this could be confusing for some users. But if we show something for the missing audio track, I don't think the user should be able to interact with that section. Therefore I resume:
- the checkbox for the track with no audio should be deselected
- it should not be possible to select the checkbox for the track with no audio
- instead of a placeholder image just a text like "No audio track available" could be displayed

luniki:
> > > Or why don't we just remove the checkbox and image completely? If there is no audio track we shouldn't show one. Instead we could just show a "rectangle" as a placeholder:
> > 
> > 
> > Might be because of my developer glasses, but to me that looks like the page is broken :D
> 
> You may be right, this could be confusing for some users. But if we show something for the missing audio track, I don't think the user should be able to interact with that section. Therefore I resume:
> 
>     * the checkbox for the track with no audio should be deselected
> 
>     * it should not be possible to select the checkbox for the track with no audio
> 
>     * instead of a placeholder image just a text like "No audio track available" could be displayed
> 
>     * video and audio boxes should have the same size and be aligned

This patch adresses all of the todo items: 
[0001-Align-audio-tracks-and-add-descriptive-text-when-mis.patch.txt](https://github.com/user-attachments/files/15837086/0001-Align-audio-tracks-and-add-descriptive-text-when-mis.patch.txt)

Screenshot:
Screenshot 2024-06-14 at 12-54-39 Opencast Editor

We could hide the checkbox if the audio is missing:
Screenshot 2024-06-14 at 12-54-59 Opencast Editor

snoesberger:
Thanks for the improvements! I can confirm that the following points are now fixed:

- [x] the checkbox for the track with no audio should be deselected
- [x] it should not be possible to select the checkbox for the track with no audio
- [x] instead of a placeholder image just a text like "No audio track available" could be displayed
- [x] video and audio boxes should have the same size and be aligned

The problem where a [placeholder image is displayed instead of the waveform](https://github.com/opencast/opencast-editor/pull/1230#issuecomment-2152191793) is still open:
grafik
@Arnei have you had a chance to look at it yet?

Arnei:
> The problem where a https://github.com/opencast/opencast-editor/pull/1230#issuecomment-2152191793 is still open:

This is a result of a more general problem, where the editor does not put too much effort (read: pretty much no effort) into associating waveform(s) with their respective videos. This was fine originally, as other views beside the track selection only ever required one waveform and it did not matter which video it came from.

Fixing this properly would require a bit of rewriting, also in how we send waveforms from the backend to editor. Not sure if this should be fixed in this PR, or a separate one.

Arnei:
ToDo Notes:
The remaining ToDos are mostly in the backend. I'll still document them here in the frontend, because this is the "main" issue where discussion happens.
- Community Workflows can't handle Audio Only. There is an open PR that intends to solve that, but that seems to be running into it's own problems: https://github.com/opencast/opencast/pull/5856
  - Alternatively, we could add a configuration option here in the frontend to disallow Audio Only track selections by default.
- Backend cannot do audio muxing. The `track-selection` WOH that is normally used for handling audio for track selections does not support this.
  - Alternatively, don't do audio muxing? From what I've heard, most people use duplication anyway?
- Backend cannot handle 3+ videos. The `track-selection` WOH explicitly only supports two videos. Supporting more would require a rewrite of the WOH.
  - Alternatively, for now, we could disable the track selection in the frontend if there are 3+ videos.

snoesberger:
> While these "solutions" are not what exactly what we wanted, they do accurately reflect the current state of the Opencast backend and workflows. Furthermore, solving these ToDos "properly" (make OC workflows handle audio only, make select-tracks WOH handle +3 videos, implement audio muxing) is a lot of work that likely will not happen anytime soon, and I would hate to see this PR get blocked because of that. Even if this PR does not achieve all the things we set out to do, it still constitutes a considerable improvement. I would therefore like to see it merged, and create follow-up issues for the "proper" solutions.

I completely agree with you. I also prefer a first version with the basic functionalities that we can already use. The rest of the customisation can be done in separate PRs.

snoesberger:
@lkiesow do you also agree with the procedure described by @Arnei in [this comment](https://github.com/opencast/opencast-editor/pull/1230#issuecomment-2194213357)?
I'd love to see this PR merged, because then we could add the translations to Crowdin and use this feature properly in production.

snoesberger:
We have now been running this PR in production for almost two months without any problems. From our user point of view (without being able to judge the code quality), this PR could be merged.

src/main/TrackSelection.tsx
```diff
@@ -1,73 +1,198 @@
...
 import {
+  selectCustomizedTrackSelection,
+  selectVideos,
+  selectWaveformImages,
+  setAudioEnabled,
+  setCustomizedTrackSelection,
+  setVideoEnabled,
+} from "../redux/videoSlice";
+import {
+  BREAKPOINT_MEDIUM,
```
JulianKniephoff, line 17:
```suggestion
  BREAKPOINTS,
```

src/main/TrackSelection.tsx
```diff
@@ -77,155 +202,244 @@ const Header: React.FC = () => {
...
-
-  const headerStyle = css({
-    fontWeight: "bold",
-    fontSize: "larger",
-    color: `${theme.text}`,
-    "&:first-letter": {
-      textTransform: "capitalize",
+    opacity: track.video_stream.enabled ? "1" : "0.5",
+    maxWidth: `${imagesMaxWidthMedium}px`,
+    [`@media (min-width: ${BREAKPOINT_MEDIUM}px)`]: {
```
JulianKniephoff, line 246:
```suggestion
    [`@media (min-width: ${BREAKPOINTS.medium}px)`]: {
```

src/main/TrackSelection.tsx
```diff
@@ -77,155 +202,244 @@ const Header: React.FC = () => {
...
+  const disabled = !customizable ||
+    (!settings.trackSelection.atLeastOneVideo && track.audio_stream.enabled && enabledCount === 1);
+
+  const imgStyle = css({
+    height: "54px",   // Keep height consistent in case the image does not render
+    width: "100%",
+    filter: `${theme.invert_wave}`,
+    color: `${theme.inverted_text}`,
+    maxWidth: `${imagesMaxWidthMedium}px`,
+    [`@media (min-width: ${BREAKPOINT_MEDIUM}px)`]: {
```
JulianKniephoff, line 307:
```suggestion
    [`@media (min-width: ${BREAKPOINTS.medium}px)`]: {
```

JulianKniephoff (dismissed):
CI fails. This should make it work.

cf. https://github.com/opencast/opencast-editor/commit/60939d9eb5c82ba8ca2b8fdac54ca2356c3ffb47#diff-1a831a072af0a35e8057bc0f12c7427b0eb10fdf4bc4eff1ed3f308e23af7a7bL38-R43

marwyg:
I overlooked some GitHub warnings in my last commit. Now everything should be fixed. @Arnei I created a PR against your Branch again: https://github.com/Arnei/opencast-editor/pull/271. Beside my successful tests, this feature is already running on a production system (since last summer I think), so it seems to work fine.

I also took a look on the code. Besides the GitHub warnings (that should be hopefully dealt with by now) , the code looks good to me. 

I think the PR can be merged now.

PR #1260, opened 2024-01-19
Enforce eslint no-explicit-any

Arnei:
Adds more typing, but also adds a few eslint exceptions. I tried to keep the exceptions to types were it makes sense to keep `any` (i.e. third party libraries returning it or just cases were we don't care about the contents of a variable). This may be controversial though.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,8 +4,7 @@ module.exports = {
     "react-app/jest",
   ],
   rules: {
-    // Currently 120 warnings.
-    "@typescript-eslint/no-explicit-any": "off",
+
   },
   overrides: [
     {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,9 +3,6 @@ module.exports = {
     "@opencast/eslint-config-ts-react",
     "react-app/jest",
   ],
-  rules: {
-
-  },
   overrides: [
     {
       files: ["./*.js"],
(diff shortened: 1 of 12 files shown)


.eslintrc.js
```diff
@@ -4,8 +4,7 @@ module.exports = {
     "react-app/jest",
   ],
   rules: {
-    // Currently 120 warnings.
-    "@typescript-eslint/no-explicit-any": "off",
+
   },
```
geichelberger, line 8: Does it have to be empty, or can we remove that?

src/redux/subtitleSlice.ts
```diff
@@ -252,8 +252,45 @@ export const selectHasChanges = (state: { subtitleState: { hasChanges: subtitle[
...
+//         }
+//       }
+//     }
+
+//     dispatch(setCurrentlyAt(milliseconds));
+//     if (triggered) {
+//       dispatch(setPreviewTriggered(true));
+//     }
+//   };
+// }
```
geichelberger, line 291: Is that a draft?

geichelberger (changes requested):

.eslintrc.js
Arnei: Can remove that.

src/redux/subtitleSlice.ts
Arnei: Iirc I was testing out function definitions, this can go now.

geichelberger (approved):
LGTM

PR #1261, opened 2024-01-19
Fix web console error message about xmlnsSvg

Arnei:
Starting the editor will prompt an error message about spelling in our opencast-editor.svg. This complies with the solution suggested by the error message, also because I'm fairly certain that we do not care about this parameter at all.


geichelberger (approved):
LGTM

The error message is gone; the logo is still displayed.

PR #1263, opened 2024-01-19
Add memoization to "selectVideos" selector

Arnei:
On behalf of a web console warning in the track selection view, this adds memoization to the "selectVideos" selector. This seems reasonable if just for the fact that track availability should never ever change.


geichelberger (approved):
Seems reasonable

PR #1264, opened 2024-01-19
Fix tooltip not showing on subtitle select button

Arnei:
Satisfies a material-ui quirk concerning `ThemedTooltip` and button. Will now correctly show a tooltip on the subtitle select button even if it is disabled.


geichelberger (approved):
LGTM

PR #1265, opened 2024-01-29
Replace CRA with Vite

### LINKED ISSUE #1253: Get rid of create-react-app
This project uses Create-React-App. Create-React-App has been in-officially deprecated for quite some time now (the last release was over two years ago and active development has ceased since then). This will eventually lead to this project running into dependency versioning issues. For example, we're already running a typescript version unsupported by CRA.

In the interest of keeping this projects dependencies up-to-date, we should eventually get rid of CRA. This issue may serve as a discussion hub on how to best move forward.

Arnei:
One way to resolve #1253.

Demonstration of replacing create-react-app with vite. Got this working in a couple hours.

Vite was chosen because:
- Straightforward migration
- Relatively similar set of opinions as CRA
- Optimized for single page applications like ours
- Popular

Fairly happy with this so far, if only for the fact that it let's us get rid of some old tech debt. And vite is indeed miles faster in development than CRA. Also gets the size of our build folder down from 9,8M to 1,7M.

ToDos:
- Playwright broke, fix it.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,6 @@
 module.exports = {
   extends: [
     "@opencast/eslint-config-ts-react",
-    "react-app/jest",
   ],
   rules: {
     // Currently 120 warnings.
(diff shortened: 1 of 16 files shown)


geichelberger (approved):
LGTM

The build is still in the same folder, and Vite is working as expected.

PR #1266, opened 2024-01-31
Fix subtitle select page crashing

Arnei:
Putting our selectors into ´createSlice` caused one of the selectors to be called incorrectly. This fixes that.


geichelberger (approved):
LGTM

It crashed; after applying the patch, it didn't.

PR #1267, opened 2024-01-31
Fix subtiles not saved for new subtitles 

### LINKED ISSUE #888: Editor does not warn if you throw away subtitles
If you change metadata and leave the page without saving, the editor warns you.
If you edit subtitles and leave the page, it silently throws away your changes.
The editor should warn the user.

Arnei:
Creating a new subtitle from scratch was broken, because the text in the cues was not saved to Opencast.
This fixes that.

Includes #1266. 

Should also fix #888.


geichelberger (approved):
LGTM

PR #1297, opened 2024-03-18
Set vite base

geichelberger:
Vite does not honor the PUBLIC_URL environment variable; we need to set the base via that variable to support building versions with a path different from the root.


Arnei (approved):

geichelberger:
Do you have any idea why the playwright tests are failing?

Arnei:
Not really, no. I kinda hoped it would be chromium just having a hiccup or something, and that the test would work if I simply reran them the next day, but that does not seem to be the case.

geichelberger:
With my last commit, the playwright tests are working locally; there was a duration mismatch. The tests are using the develop instance, and I guess https://github.com/opencast/opencast/pull/5669 changed the metadata :smile:

Arnei:
Can confirm tests running locally (Quite unexpected to get an actual duration value in the metadata :D). Still running into timeouts in gh actions though.

geichelberger:
The commit [Remove unknown fallback language](https://github.com/opencast/opencast-editor/pull/1297/commits/cae00400ab85b8d1d965b6d342cd7028991dc582) should solve the problem that sometimes that langstrings are not shown. At least after that change, I was not able to reproduce it.

However, that did not solve the problem with the playwright tests; it was caused by a problem with the constructed settings URL that needed to be corrected.

Arnei (approved):

Arnei:
Amazing find!

PR #1310, opened 2024-04-04
Add lazy loading of lang files

geichelberger:
Only the currently used language is loaded instead of all language files from the beginning. A new lazy loading backend loads the lang files as a dynamic import.

This should fix the bug that sometimes the language keys are missing.
grafik


Arnei (commented):
Works. Can't really confirm whether this fixes the issue with languages not loading sometimes as I cannot consistently reproduce it, but at the very least it does not break anything obvious :D

That being said, does renaming the locales break our crowdin workflow?

geichelberger:
> That being said, does renaming the locales break our crowdin workflow?

I missed that, the locales.json is generated automatically, this also resolves the issues of populating the array. I will change that.

geichelberger:
I changed how the crowdin translation array/map is generated by replacing the `locales.json` with an `lngs-generated.ts` file. The locales now have the same name as before, and the generation script creates a map of the country codes and the languages.

Arnei (approved):
Can't really test this, but in general this looks solid to me (if maybe a little convoluted, but that's more i18ns fault than anything).

geichelberger:
I changed the script path to use the absolute path with $GITHUB_WORKFLOW because the action uses a working directory.

KatrinIhler:
Discussed and approved, change sounds reasonable -> merging.

Issue #1131, opened 2023-08-18
UX/UI: Clarify confusion around "Save changes" vs. "Start processing"


oas777:
Testing at ETH showed the difference between the two and their specific functionalities are not self-explanatory, despite the description. We would appreciate a discussion around this.

oas777:
#https://github.com/elan-ev/opencast-studio/issues/1061

lkiesow:
Our users ran into the same problem. “Saving” vs “Processing” is really confusing. Especially if you just want to update the metadata. At least that we can hopefully improve once the auto-update of metadata is finished which is under development right now.

oas777:
Can you elaborate on this feature? While I can see it help in certain situation, it will not necessarily solve this UX/UI issue.

oas777:
Needs further discussion, maybe input by designer (Lisa).

Arnei:
Agreed, this totally unintuitive. However, I also find it really difficult to abstract from the Opencast context. A designer input would be greatly appreciated.

oas777:
Cf. https://github.com/elan-ev/opencast-studio/issues/1099 also.

lkiesow:
As a first step, let's overhaul the naming.
@kai-wng-uos, can you describe what we discussed in more detail?

A few notes from our discussion:

- Start processing also saves changes
- Start processing should mention publication instead
- Go through labels and wording in general

PR #1311, opened 2024-04-15
Change wordings in finish menu

dennis531:
Change the the wording of the "Start processing" button to "Save and publish changes" to better differentiate the buttons. Additionally, this should highlight that videos can be published with this button.

I haven't changed the existing translations, as I don't know what to do with them.

Related #1131

### DIFF
--- a/src/i18n/locales/de-DE.json
+++ b/src/i18n/locales/de-DE.json
@@ -49,7 +49,7 @@
     },
     "finishMenu": {
         "save-button": "Änderungen speichern",
-        "start-button": "Verarbeitung starten",
+        "start-button": "Änderungen speichern & veröffentlichen",
         "discard-button": "Änderungen verwerfen"
     },
     "save": {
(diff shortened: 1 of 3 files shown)


ziegenberg:
I don't know if that change is desirable. The workflow may or may not publish the video. How about "Save and start processing"?

dennis531:
> I don't know if that change is desirable. The workflow may or may not publish the video. How about "Save and start processing"?

I had the same thoughts when I changed the texts. I agree that we should keep "processing" and append "save" to the text.

KatrinIhler:
Change looks fine, but i wonder how we can trigger a re-translation of this. Maybe we should rename the translation key (append `new` or something)?

ziegenberg:
Shouldn't Crowdin handle those cases? The English string has changed, so it comes up again for all translators.

KatrinIhler:
> Shouldn't Crowdin handle those cases? The English string has changed, so it comes up again for all translators.

I don't know if that's how it works? If it is, that'd be great.

ziegenberg:
Maybe @gregorydlogan or @lkiesow, who know how Crowdin works, want to chime in. We should document "How to handle translation" afterwards.

KatrinIhler:
From the dev meeting: We assume that changing the English translation string will prompt re-translation into the other languages in Crowdin. :)

gregorydlogan:
Will clarify in the docs as a separate PR, but yes that's my understanding.  Changing the translated (non-en_US) strings is frowned upon because they will blindly be overwritten by the Crowdin bot and this can also cause merge conflicts going upwards towards Develop.

dennis531:
@gregorydlogan Thanks for the clarification. I reverted the german translations.

PR #1316, opened 2024-04-19
Subtitle editor: Allow to change end time of segment in timeline by drag and drop

### LINKED ISSUE #1314: Subtitle editor: change length of subtitle segment in the timeline
It would be nice if a user could adjust the length of a subtitle segment in the timeline using drag and drop. More precisely the user would click on the end (or the start) of the segment in the timeline and drag it to the desired position:
grafik

snoesberger:
In the subtitle editor timeline, it was only possible to change the start time of a segment by drag and drop. This change makes it possible to change the end time in the same way.

Resolves #1314

### DIFF
--- a/src/main/SubtitleTimeline.tsx
+++ b/src/main/SubtitleTimeline.tsx
@@ -287,24 +287,13 @@ const TimelineSubtitleSegment: React.FC<{
   const onResizeAbsolute: ResizableProps["onResize"] = (_event, { size, handle }) => {
     // Possible TODO: Find a way to stop resizing a segment beyond 0ms here instead of later
     let newLeft = absoluteLeft;
-    let newTop = absoluteTop;
-    const deltaHeight = size.height - absoluteHeight;
-    const deltaWidth = size.width - absoluteWidth;
-    if (handle[0] === "n") {
-      newTop -= deltaHeight;
-    } else if (handle[0] === "s") {
-      newTop += deltaHeight;
-    }
+
     if (handle[handle.length - 1] === "w") {
-      newLeft -= deltaWidth;
-    } else if (handle[handle.length - 1] === "e") {
-      newLeft += deltaWidth;
+      newLeft -= size.width - absoluteWidth;
     }
 
     setAbsoluteWidth(size.width);
-    setAbsoluteHeight(size.height);
     setAbsoluteLeft(newLeft);
-    setAbsoluteTop(newTop);
   };
 
   // Update redux state based on the resize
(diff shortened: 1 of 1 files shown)


KatrinIhler:
It's a bit unfortunate that I can't test this just because the ICLA hasn't been processed yet. :sweat_smile:

snoesberger:
I filled in the ICLA a week ago, but haven't received a confirmation yet.

lkiesow:
> It's a bit unfortunate that I can't test this just because the ICLA hasn't been processed yet. 😅

Maybe to avoid future confusion about this, a few words about the test deployment, which absence I think you are complaining about. First, you can always just clone this and run `npm start` locally.

Second, I think here is a misconception that the failing ICLA test has something to do with the missing test deployment. Since that is deploying code, for security reasons, it's only doing that for known developers. To be more precise, it's limited to people in the team `Developers` of the Opencast organization here on GitHub:

https://github.com/opencast/opencast-editor/blob/fbae57c560a48ba997aa9cf10de3540e3f1d9973/.github/workflows/update-test.yml#L11-L12

@gregorydlogan or I can easily add more people to that group. If you want access or want someone else to have access, just let us know. @snoesberger, I've invited you to that team.

lkiesow (changes requested):
Tried this and found a bug. It seems like you get two input fields for subtitles whenever you start a new subtitle:


https://github.com/opencast/opencast-editor/assets/1008395/85e01d0c-a1a7-4ef7-b95a-1b4cbf81afb3

KatrinIhler:
> Second, I think here is a misconception that the failing ICLA test has something to do with the missing test deployment. Since that is deploying code, for security reasons, it's only doing that for known developers. To be more precise, it's limited to people in the team `Developers` of the Opencast organization here on GitHub:

Ah, thx, I did not know this! Makes sense.

snoesberger:
> Tried this and found a bug. It seems like you get two input fields for subtitles whenever you start a new subtitle:
> opencast-editor-subtitle-two-input-fields.mp4

Thanks for testing, Lars. This is strange, I can reproduce the behaviour but only when I start the editor in development mode. If I build the editor and start the new build, only one input field is added. And I have the same behaviour when I do the same for the main branch, so it looks like this problem hasn't been added by the changes in this PR.

Arnei (approved):
Works and simplifies the code to boot, nice.

Can confirm that the concern Lars brought up is unrelated to this pull request, as it is already happening on main.

snoesberger:
I'd love to see this PR merged. Or is there still something preventing it from being merged?

Arnei:
I don't see any blockers anymore, thanks for the reminder.

Issue #1283, opened 2024-02-23
Jump to next previous controls


lkiesow:
When dealing with small sections, it would be really helpful to add two new playback controls to jump to the next and previous segment.

lkiesow:
Fixed by #1317

PR #1317, opened 2024-04-22
Controls to jump to next or previous segment

dennis531:
Add the mentioned controls in #1283. 

The controls are deactivated on the subtitle interface, as in my opinion the function is not needed here.

Screenshot with the controls:
image


lkiesow:
I found two problems with this:

1. When I'm in the middle of a segment and hit “jump back”, I would expect the end up at the start of the segment, not at the start of the previous segment. Or in other works, I would expect to jump to the next or last cut mark
2. I cannot jump while the video is playing


https://github.com/opencast/opencast-editor/assets/1008395/4bdfb5aa-0497-4d65-83de-a08eab946efc

kai-wng-uos:
I agree with @lkiesow 
that is also how it works in all media players I have used. maybe we should change the wording to "zum letzten Schnitt springen"

dennis531:
> I found two problems with this:
> 
>     1. When I'm in the middle of a segment and hit “jump back”, I would expect the end up at the start of the segment, not at the start of the previous segment. Or in other works, I would expect to jump to the next or last cut mark
> 
>     2. I cannot jump while the video is playing
> 
> 
> opencast-editor-jump-segment.mp4

I fixed the mentioned problems. 

to 1: I added a 3 second threshold at the beginning of the segment as it was not possible to jump back to the next video when the playback is active. For the duration, I oriented myself on spotify's "jump back" function. 

>  that is also how it works in all media players I have used. maybe we should change the wording to "zum letzten Schnitt springen"

I think we should change the wording too. Alternatively, we could provide a more general label like "Jump back" without explicitly mention the target. Especially with the treshhold, I'm not sure if your suggestion still fits.

kai-wng-uos:
looks good to me now, Thanks @dennis531! As for the wording: I think a simple back/ forward might suffice. Users are already used to this wording with other media players.

KatrinIhler:
Conflicts. :(

lkiesow:
Looks good. I have a suggestion for a minor improvement for the 3 seconds threshold, but I'll file a separate pull request for that.

PR #1318, opened 2024-04-25
Add context menu for cutting actions

### LINKED ISSUE #943: Add right click menu for cutting

dennis531:
Displays a context menu over the cutting timeline when the user clicks on it with the right (secondary) mouse button.

Closes #943


lkiesow (approved):
Looks good to me

kai-wng-uos:
I really like this change!

ziegenberg:
Shouldn't we only change the en-us.json locale file because Crowdin does the translation?

dennis531:
I reverted the tranlation files as suggested in https://github.com/opencast/opencast-editor/pull/1311#issuecomment-2086343395.

lkiesow:
Shouldn't really matter since they would just be overwritten by the Crowdin changes. But it's reverted now, so nevermind.
Merging.

PR #1329, opened 2024-05-06
Dynamic skip threshold

lkiesow:
The hard-coded threshold of 3s for skipping to the previous segment instead of skipping to the start of the segment can be too long for short videos, where visually, you are far away from the actual cut mark.

Screenshot from 2024-05-06 19-03-42

This patch fixes the problem by making this a dynamic threshold based on the video duration with a minimum of 0.5s (guarding against a non-set video duration) and a maximum of 3s.

### DIFF
--- a/src/redux/videoSlice.ts
+++ b/src/redux/videoSlice.ts
@@ -150,14 +150,18 @@ const videoSlice = createSlice({
     jumpToPreviousSegment: state => {
       let previousSegmentIndex = state.activeSegmentIndex - 1;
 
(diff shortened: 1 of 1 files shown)


dennis531:
Thanks for the fix! It looks good to me.

Note: The maximum threshold of 3 seconds is reached with a video duration of 5 minutes. Otherwise it will be shorter.

ziegenberg (approved):

lkiesow:
Thanks for the review. Merging this so I can prep the new release…

PR #1336, opened 2024-05-08
Make editor logo changeable

ciegler:
Makes the editor logo changeable similar to the studio logo.


lkiesow:
I'm not sure if we really want this. I think it's a good thing for Opencast as a project to keep the branding intact and therefor keep the logo in the tool. Any particular reason to replace this?

KatrinIhler:
From the Dev meeting: There were general thoughts on rebranding, but nobody was strongly against this specific pull request.

lkiesow:
You added the logo to somewhere else, but it's also still at the old location, isn't it? Can we move that instead?

ciegler:
Yes, I will do that. I also saw that I cannot add a label to the PR and it looks like a proper label is required. Is there a way in which I can add a label?

majosch:
As member of the institution (University of Vienna) that requested this feature I just wanted to add some notes about the motivation - I couldn't speak up in technical meeting because I was in a crowded tram with headphones only.

Personally I'm no big fan of relabeling, because I think it is misleading for users, causes unnecessary overhead and you're running in inconsistencies. But it's not my call.

On the other hand I'm running three instances with multiple tenants and I use different color schemes/logos to be able to identify the actual instance/tenant at a glance.
Doing this on the webserver is the most convenient way.

marwyg:
I build this version locally and it seems to work:
grafik

I changed the window size (horizontally and via zoom) and I switched between all light/dark modes, everything seem to work (it also fixes a bug, where the logo changes it size when you switch between modes). The code also looks fine to me.

Arnei (changes requested):
Please fix the checkstyle issues.

This changes the logo CSS. The logo now looks slightly different from before, making it less similar to Studio than before? The logo is shifted slightly to the right, and has less contrast in dark mode.

marwyg:
I also noticed, that the logo is styled slightly different, but I think, these changes are intentionally.

ciegler:
Thanks for the comments, since this was the first project I was working on I completely forgot to check the Logo's appearances in different Themes. The contrast in darkmode is now similar to the contrast in Studio and the slightly shift is also gone.

Arnei (changes requested):
Vite still reports three checkstyle issues.
```
   96:1  warning  Expected indentation of 4 spaces but found 6  indent
  106:1  warning  Expected indentation of 4 spaces but found 6  indent
  107:4  warning  Missing semicolon                             semi
```

The CSS looks good to me now. Since Martin said this works, I believe this can be merged once the checkstyle stuff is out of the way.

Arnei (approved):

PR #1337, opened 2024-05-13
Make videos in cutting view larger

Arnei:
I have heard feedback that the video(s) in the cutting view are too small since the redesign. Therefore, this PR makes them slightly larger, but not as screen filling as before. Since this change focuses on design first and foremost, discussion would be appreciated on whether we want larger video(s) and if this might inconvenience current users somehow.

Current:
Bildschirmfoto vom 2024-05-13 12-15-46

Larger:
Bildschirmfoto vom 2024-05-13 13-49-06


KatrinIhler:
From the Dev meeting: This causes scrolling for quite a few people, which is kind of annoying. :sweat_smile:

lkiesow:
Looking at your first image, I agree that there is unused space we can use.

But looking at this on my 16:9 display, this pull request now causes the controls to be hidden and I have to scroll within the editor. That's really not great:

Screenshot from 2024-05-14 17-54-31

↑↑↑ controls are hidden here. Not great.

Arnei:
So ideally we ensure that there is no scrolling on the Cutting page (at least for common resolutions), while still having the videos take up more space? I can look into that.

snoesberger:
The videos look much smaller in Safari than in Firefox, for example.

Safari:
grafik

Firefox:
grafik

By the way, even now, without this PR, the videos are much smaller in Safari than they are in other browsers.

Arnei:
The difference is Safari is likely related to another issue that was very recently fixed here https://github.com/opencast/opencast-editor/pull/1383

PR #1341, opened 2024-05-28
Specify version of comment action

lkiesow:
Instead of just specifying the main branch, this pull request specifies the latest major version of thollander/actions-comment-pull-request to use in GitHub Actions. We don't know how this action changes and tracking the default branch may cause this to break at any time. We can always update using Dependabot.


ziegenberg:
Sounds reasonable.

Issue #1282, opened 2024-02-23
Thumbnail Editor in simpleMode looks weird


majosch:
Just built the latest master and deployed as standalone.

If the thumbnail editor is configured with `simpleMode = true` the buttons in the "Change thumbnail here"-canvas(?) are aligned under the preview image:

![Bildschirmfoto vom 2024-02-23 11-21-35](https://github.com/opencast/opencast-editor/assets/16731250/994fac17-fad4-45d3-ae36-fc2af5cab67e)

Configured with `simpleMode = false`:

![Bildschirmfoto vom 2024-02-23 11-32-13](https://github.com/opencast/opencast-editor/assets/16731250/367df367-4516-425c-8017-bab9ec90e5c4)

I guess the buttons should be aligned the same way in both thumbnail editor modes?

PR #1365, opened 2024-05-30
Fix initial appearance of thumbnail view

Arnei:
We want the boxes for the thumbnails and the videos to be roughly the same size, and display the buttons in the thumbnail boxes on the right if space allows. This was not happening when first clicking on the thumbnail view, wasting space and giving it a generally not so great look. This changes some max widths around to fix that.

Potentially a fix for #1282.

Before
Bildschirmfoto vom 2024-05-30 09-42-10

After
Bildschirmfoto vom 2024-05-30 09-42-47

### DIFF
--- a/src/main/Thumbnail.tsx
+++ b/src/main/Thumbnail.tsx
@@ -252,7 +252,7 @@ const ThumbnailTableRow: React.FC<{
     // The "+40" comes from padding that is not included in the "getWidth" function
     generateRef.getWidth() + 40 :
     // Random default
-    440;
+    740;
 
   const renderPriority = (thumbnailPriority: number) => {
     if (isNaN(thumbnailPriority)) {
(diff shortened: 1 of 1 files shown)


snoesberger:
Looks good to me in various browsers (Firefox, Chrome, Safari) with one small flaw. If you have a dual stream recording and a small browser window, the thumbnail placeholder does not scale. If you have a single stream recording, the scaling of the thumbnail placeholder works fine:
grafik
grafik


One other remark: in Safari the preview videos are smaller than in other browsers, but this is already fixed by this PR https://github.com/opencast/opencast-editor/pull/1383

gregorydlogan:
@Arnei are you able to fix that minor issue here, or should we open a new PR later for it?

Arnei:
I do not think I will be able to fix that minor issue here iirc. Hopefully it is not a problem for many users.

PR #1367, opened 2024-05-30
Add cue at current time for subtitles

### LINKED ISSUE #1313: Subtitle editor: insert a new subtitle segment at the specific playtime
It would be nice if a user could insert a new subtitle segment at the current time of the player. The start time of the subtitle segment should be the current position on the timeline:

grafik

One could imagine using a keyboard shortcut, double-clicking on the timeline/waveform or using an additional UI element.

Arnei:
Adds a hotkey (Shift+Alt+e) to the subtitle editor. It allows the user to add a new subtitle segment, based on the current time of the video.

Resolves #1313.


snoesberger:
Thanks for that. There is one small issue, every time I use a keyboard shortcut to add a subtitle segment, the special character corresponding to the shortcut is added to the new segment. I would expect it to just add a new empty segment.

This is not only true for Shift+Alt+e but also for Shift+Alt+q and Shift+Alt+a.

geichelberger:
@snoesberger I could not reproduce it. Is it still an issue for you?

snoesberger:
> @snoesberger I could not reproduce it. Is it still an issue for you?

Yes, it is still a problem, I just tried again. It depends on the keyboard language you are using. If you're using a standard German keyboard layout (de), you won't see the problem. With a Swiss-German (de-ch) or US layout I'm able to reproduce the problem in different browsers.

src/main/SubtitleTimeline.tsx
```diff
@@ -94,6 +108,11 @@ const SubtitleTimeline: React.FC = () => {
     () => setKeyboardJumpDelta(keyboardJumpDelta => Math.max(keyboardJumpDelta / 10, 1)),
     {}, [keyboardJumpDelta]
   );
+  useHotkeys(
+    KEYMAP.subtitleList.addCue.key,
+    () => addCue(currentlyAt),
+    {}, [currentlyAt]
```
gregorydlogan: This line throws an error at build time:
```
/home/greg/opencast/opencast-editor/src/main/SubtitleTimeline.tsx                                                                             
  114:22  warning  Missing trailing comma  comma-dangle                                                                                       
                                                                                                                                              
✖ 1 problem (0 errors, 1 warning)                                                                                                             
  0 errors and 1 warning potentially fixable with the `--fix` option.                                                                         
                                                                                                                                              
ESLint found too many warnings (maximum: 0).  
```

gregorydlogan (changes requested):

src/main/SubtitleTimeline.tsx
Arnei: Thanks, should be fixed now.

PR #1376, opened 2024-06-03
Fix editor not playing audio due to video order

Arnei:
If the editor has multiple videos, it will consider one of them the "primary" video, from which it will play back audio. Even if that "primary" video does not have any audio.
This changes the logic so that the first video with audio the editor comes across is made the "primary" video, so that there should always be audio playing if audio is available.

### DIFF
--- a/src/main/VideoPlayers.tsx
+++ b/src/main/VideoPlayers.tsx
@@ -21,6 +21,7 @@ import {
   setJumpTriggered,
   selectJumpTriggered,
   setCurrentlyAt,
+  selectVideos,
 } from "../redux/videoSlice";
 
 import ReactPlayer, { Config } from "react-player";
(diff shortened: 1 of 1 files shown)


snoesberger:
I tested it with a dual stream recording where I had problems before. I can now hear the audio in the cutting section of the editor, but I still have the problem in the preview player of the subtitles section.

Arnei:
The video player in the subtitle section should be completely unaffected by this change.

The subtitle section only displays one video at a time and thus, only the audio from the currently selected video is played. If the video has no audio, then no audio will be played. Not sure if this can be fixed, but this PR at least is not trying to do anything about it, so I would consider it a separate issue.

snoesberger:
> The video player in the subtitle section should be completely unaffected by this change.
> 
> The subtitle section only displays one video at a time and thus, only the audio from the currently selected video is played. If the video has no audio, then no audio will be played. Not sure if this can be fixed, but this PR at least is not trying to do anything about it, so I would consider it a separate issue.

You are right, this makes sense. But I can see how it might be confusing for some of our users. But for now, let's leave it as you suggested (and rethink if there's a better way to handle this, I have opened the following issue for that: https://github.com/opencast/opencast-editor/issues/1380).

snoesberger:
I'd love to see this PR merged. Or is there still something preventing it from being merged?

Arnei:
This still needs a formal review I'd say.

snoesberger (commented):
As GitHub Actions has already discovered, selectVideoURL is no longer used and could be removed from the import.
Otherwise, the code changes look reasonable to me, and my tests with different dual stream recordings were successful.

PR #1383, opened 2024-06-10
Fix aspect ratio video box size in Safari

### LINKED ISSUE #1382: Safari: Boxes around videos do not fit the videos
In Safari, the boxes around the videos seem to always fill the screen, even if the videos themselves do not.

Tested with whatever Safari version is on MacOS 17.3 right now.

grafik

ferishili:
This PR fixes #1382 

### Problem
The problem is simply the lack of support in Safari that it could not recognize and allocate height based on aspect ratio.
The reason behind it, is because Safari support for aspect ratio was introduced in v15, and it is far behind the other browsers.

### Solution
In order to tackle that, this PR just introduces a wrapper around the React video player, by allowing it to have the height of 100% set for each video! That means now the height is not more a burden on browser to calculate based on aspect ratio!
The wrapper on the other hand is responsible to occupy the width and height designed for this section and let the video just take up the space it needs!

NOTE: it not only works for dual video, but the single video is covered too!

### To test
- you would need to have Safari to see the difference!
- patch this PR (locally, if you will)
- npm start (locally, if you will)
- Make sure you test it with single and dual videos in cutting page!

### DIFF
--- a/src/main/VideoPlayers.tsx
+++ b/src/main/VideoPlayers.tsx
@@ -413,7 +413,7 @@ export const VideoPlayer = React.forwardRef<VideoPlayerForwardRef, VideoPlayerPr
       // For multi videos, last from right side, sitting on start
       ...(last && !first) && { justifyContent: "start" },
 
-      // For multi videos, in between, occupy only the content!
+      // For multi videos, in between, fit content and center!
       ...(!first && !last) && { justifyContent: "center", flexBasis: "fit-content" },
     });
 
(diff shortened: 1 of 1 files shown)


Arnei:
And now I'm gonna be the mean person that reminds everyone that we technically also support more than two videos.
Bildschirmfoto vom 2024-06-10 11-43-12
While three videos also looks good, it looks different from two videos in that the videos are now spaced out instead of grouped up in the center. I think it would be preferrable if the three videos were also grouped up in the center. What do you think?

ferishili:
Well, good to know that more than 2 videos are also supported, let me fix that.

Arnei (approved):

oas777:
Late to the game, but this works for me, Safari shows the same size Chrome and Firefox display. Thanks, Farbod!

ferishili:
Great, with my pleasure.

PR #1387, opened 2024-06-12
Remove flex gap polyfill

JulianKniephoff:
According to our browserslist config, this is only needed for very old versions of mobile Safari, which we don't support anyway.

cf. https://browsersl.ist/#q=%3E0.2%25%0Anot+dead%0Anot+op_mini+all%0Anot+fully+supports+flexbox-gap

### DIFF
--- a/src/cssStyles.tsx
+++ b/src/cssStyles.tsx
@@ -4,7 +4,6 @@
 import { css, Global, keyframes } from "@emotion/react";
 import React from "react";
 import emotionNormalize from "emotion-normalize";
-import { checkFlexGapSupport } from "./util/utilityFunctions";
 import { createTheme } from "@mui/material/styles";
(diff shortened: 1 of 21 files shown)


JulianKniephoff:
It works in my version 🤷‍♀️ I don't know how much that is worth in relation to caniuse data. x)

JulianKniephoff:
Note, too, that the workaround isn't perfect. It adds a `margin` indiscriminately whereas `gap` only applies to flex or grid containers. This is actually relevant for this project since the helpers I'm removing here are indirectly used in non-flex or -grid contexts as a side effect of using other styles which happen to call them. That's a problem in and of itself, but not as easily fixed, sadly. 😅

Arnei (approved):

Arnei (approved):

Issue #1381, opened 2024-06-06
Make further use of appkit


Arnei:
By now `@opencast/appkit` has quite a few components we could and likely should make use of. I believe the following appkit components can replace and improve our current code:
- A Spinner component
- An Error Box component
- Generic button components

PR #1392, opened 2024-06-27
Make use of error boxes from appkit

Arnei:
The error boxes are small notification boxes that inform the user if part of the editor encountered a problem that does not completely crash the application. This replaces our boxes with the ones from appkit, which look way better.

Helps with #1381.

Example screenshots:
Before:
Bildschirmfoto vom 2024-06-27 12-26-11
After:
Bildschirmfoto vom 2024-06-27 12-25-49


geichelberger (approved):
LGTM

PR #1393, opened 2024-06-27
Make use of ProtoButton from appkit

Arnei:
Replaces our home-brewed buttons with the basic ProtoButtons from appkit. There is arguably more  work to be done on buttons, but this should serve as a good foundation for such future work.

Helps with #1381.

No screenshots this time, as everything should look and work the same. The benefits here are in code quality and potentially improvements in accessibility.


geichelberger (approved):
LGTM

PR #1414, opened 2024-07-09
Center final message again

Arnei:
The message after successfully starting a workflow was stuck to the top of the page. This patch moves it to the center of the page again.


lkiesow:
Ups, did miss that. Sorry.
This does seem a bit less magical and therefor hopefully less likely to break. But I don't really care either way as long as it gets fixed. Do you have a preference?

lkiesow:
Okay, closing this, closing #1467 to finish this.

PR #1416, opened 2024-07-09
Remove CRA ts definition file

Arnei:
A file used by create-react-app. Since we switched to Vite, we don't need this anymore.


geichelberger (approved):
LGTM

PR #1424, opened 2024-07-17
Fix opencast editor logo

geichelberger:
Handles the static asset `opencast-editor.svg` according to the vite documentation.

https://vitejs.dev/guide/assets#the-public-directory

### DIFF
--- a/src/main/Header.tsx
+++ b/src/main/Header.tsx
@@ -92,7 +92,7 @@ function Header() {
 }
 
 const LogoPicture: React.FC = () => {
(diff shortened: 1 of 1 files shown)


majosch:
On my testing instance (15-SNAPSHOT) I've recently deployed Opencast including an editor build including this PR. Path to editor looks good now:

grafik

Arnei (approved):
And I can also confirm that this fixes the issue that caused https://github.com/opencast/opencast/pull/6030. So I dare say this should be good to go.

PR #1467, opened 2024-10-08
Properly center and space end screen

lkiesow:
This patch fixes the issue that the text and the check mark icon of the end screen is pushed to the top of the end space with basically no spacing. This is caused by an incorrect CSS calculation.

---

Before:
Screenshot from 2024-10-08 23-26-29

After:
Screenshot from 2024-10-08 23-26-14

### DIFF
--- a/src/main/TheEnd.tsx
+++ b/src/main/TheEnd.tsx
@@ -34,7 +34,7 @@ const TheEnd: React.FC = () => {
 
   const theEndStyle = css({
     width: "100%",
-    height: "calc(100% - 64px)",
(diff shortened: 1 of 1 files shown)


Arnei:
Do you want this to supercede #1414 ?

lkiesow:
Ups, did miss that. Sorry.
This does seem a bit less magical and therefor hopefully less likely to break. But I don't really care either way as long as it gets fixed. Do you have a preference?

Arnei (approved):

PR #1472, opened 2024-10-15
Unify save requests

### LINKED ISSUE #485: One request to save and start workflow
Instead of having two REST calls for saving and starting a workflow, it would make sense to use a single one instead to ensure that the process is transactional, we don't send requests in an incorrect order and so that we need less requests overall. After all, you will never start a workflow without saving the data first, starting this on old data.

Arnei:
Fixes #485.

When the user was saving (or saving and processing), we would sent two sequential POST requests, one for metadata and one for all the rest. This unifies the requests into a single one.

Requires backend changes: https://github.com/opencast/opencast/pull/6241


geichelberger (approved):
LGTM

PR #1495, opened 2024-11-25
Make behavior of forward button more intuitive

JulianKniephoff:
This patch makes the forward button jump to the end of the video if you are currently in the last segment.

The previous behavior, jumping to the **beginning** of the last segment, made it so that the forward button sometimes made you go backwards in the video. This is especially confusing if there is only one segment, in which case the forward button put you at the beginning of the video.


geichelberger (approved):
LGTM

PR #1522, opened 2024-12-19
Fix various eslint and typescript complaints

Arnei:
Fixes various small issues that don't directly impact functionality, but that eslint and typescript complain about.

### How to test this

Can be tested as usual.


geichelberger (approved):
LGTM

Issue #1523, opened 2024-12-19
Subtitle delete button is bigger than the rest


Arnei:
For whatever godforsaken reason the button for deleting subtitles is now styled differently than its accompanying buttons. These buttons should all be styled the same.

![Bildschirmfoto vom 2024-12-19 10-38-26](https://github.com/user-attachments/assets/c7abc9ed-6950-400d-945b-c7be0d7d14d0)

PR #1524, opened 2024-12-19
Fix subtitle button height

Arnei:
Fixes https://github.com/opencast/opencast-editor/issues/1523.

Some button at the top of the subtitle editor were not the right height. This patch should fix that.

Bildschirmfoto vom 2024-12-19 11-30-43

### How to test this

No special dependencies, can be tested as is. Install and look at the top row of buttons in the subtitle editor. You may need to configure `show = true` for subtitles in the `editor-settings.toml`.


geichelberger (approved):
LGTM

Issue #1565, opened 2025-03-27
Mark required metadata fields


Arnei:
A metadata catalog may have one or more required fields. These fields need to have a value, else requests to the backend will fail. The required fields should be marked (similar to how it is done in the admin ui) so that the user is aware of which fields they must fill in. Ideally we draw the users attention to those fields by e.g. highlighting them.

Different but related issue: Have the endpoint return some sensible error message. 500 without explanation is **not** helpful.

KatrinIhler:
Note: Invest at most 20h into patch, needs to be discussed further at hdn meeting if not finished by then.

PR #1566, opened 2025-03-31
Make required metadata fields more visible

Arnei:
Helps with #1565.

Overall this PR aims at making the user more aware of required fields in the metadata tab, particularly if one of those required fields is missing a value which will result in an error in the backend. The individual changes can be found in the commit history.

Does not prevent the user from clicking "Save & Process" even if there are still validation issues. Would like to add that in a different PR as a more generalized version of #1517.

### How to test this

Ideally this is tested with an Opencast you can control, so you can change which metadata fields are required in `etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg`.


wsmirnow:
`mixed_text` metadata field type seems not to work correctly. On page load missing values are accepted (see screenshot). But if you put some text there and delete it afterwards, the required-message is shown.

image

src/main/Metadata.tsx
```diff
@@ -278,7 +284,19 @@ const Metadata: React.FC = () => {
...
+  const required = (value: unknown) => {
+    let val = value;
+
+    if (value && typeof value === "object" && "submitValue" in value) {
+      val = value.submitValue;
+    }
+
+    if (value && Array.isArray(value) && value.length === 0) {
+      val = false;
+    }
```
wsmirnow, line 296: In case you fill out `mixed_text` metadata value and delete the content afterwords, the required message isn't shown as expected. The value is `['']` in this case. Lets get rid of empty strings from array before checking the length.

```suggestion
    if (value && Array.isArray(value) && value.filter(v => v).length === 0) {
      val = false;
    }
```

wsmirnow (changes requested):

src/main/Metadata.tsx
Arnei: Can't reproduce this.

[Bildschirmaufzeichnung vom 2025-05-28 10-08-07.webm](https://github.com/user-attachments/assets/96a7104a-0de9-48fc-925e-186a17c0c7ff)

Could you tell me what I'm doing wrong, or have any other tips for me on how to reproduce this? I worry that empty strings in the array point at a more fundamental problem.

wsmirnow: Sure, here is an example

[editor-required-metadata-field-on-focus-lost.webm](https://github.com/user-attachments/assets/3cd6ef33-02fe-4bad-b8ba-568d9d520a50)

The affected metadata field configuration looks like

image

wsmirnow: The delimiter configuration is not applying too. But this is a separate issue.

Arnei: Thanks for the detailed report, I forgot to consider that `mixed_text` fields may not have a listprovider defined. The latest commit 74c72b should fix the issue you were seeing.

I agree that respecting delimiter configuration is better handled in a separate issue/PR.

gregorydlogan:
@wsmirnow are you concerns resolved?

src/main/Metadata.tsx
wsmirnow: Yes, thanks.

wsmirnow (approved):

PR #1580, opened 2025-05-05
Sync 17.x state with main/develop

gregorydlogan:
This PR refactors the release process with two workflows:
- `create-release`: Creates a release tag, for use with git-reduced workflows.  Manual only trigger.
- `process-release`: Creates a release in the format of `OC_VERSION.x-YYYY-MM-DD`, files an upstream PR updating `opencast/opencast` to that release for the appropriate branch.

This PR also:
- Generates an index page for editor.opencast.org -> This requires #1581 
- Creates an editor.opencast.org/OC_VERSION page with the latest snapshot of the OC_VERSION release for the editor.

Future(?) work: 
- Document how to create the GHA PAT so that this works.  Is it *not* very clear, and will expire on 2026-05-06 :(
- Upstream documentation that this repo (and admin, and studio...) are now following the upstream development process cf branching.


gregorydlogan:
I'm assuming we want to stick with `main` here for our develop-equivalent branch?  If we're going to change it, now would be the time...

PR #1010, opened 2023-03-31
Load locally if possible

lkiesow:
This patch allows the editor to load files from the local server if possible (indicated by Opencast). This means that, for example, with the default configuration, files are no longer loaded from the admin server when the editor is accessed on presentation, causing authentication problems.

PR #1582, opened 2025-05-07
Load waveform images locally if possible

dennis531:
This patch allows the editor to load the waveform images from the local server if possible (indicated by Opencast).
The mechanism is identical to what we already use for tracks.

This patch bases on #1010


lkiesow (approved):
Works and fixes the problem that waveforms are sometimes now shown.

PR #1584, opened 2025-05-09
Changing the image-rendering of waveform images

### LINKED ISSUE #1583: Waveform looks sometimes blurry in Chrome browser
If you have long videos the waveform looks sometimes blurry in the Chrome browser. Once you zoom into the waveform the waveform looks normal again. This seems to be a Chrome issue only, in Firefox the waveform looks normal even when not zoomed in.

Chrome:

Image

Image

The same video in Firefox:

Image

Image

snoesberger:
fixes #1583

Large waveform images appear blurry in Chrome browsers. Changing the image rendering to "pixelated" for the waveforms helps to avoid this.

Example without `image-rendering: "pixelated"`:
grafik
grafik

The same waveform with this PR:
grafik
grafik

### DIFF
--- a/src/main/Timeline.tsx
+++ b/src/main/Timeline.tsx
@@ -596,6 +596,7 @@ export const Waveforms: React.FC<{ timelineHeight: number; }> = ({ timelineHeigh
   const waveformStyle = css({
     background: `${theme.waveform_bg}`,
     borderRadius: "5px",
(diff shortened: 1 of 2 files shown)


Arnei (approved):
Weird issue, weird fix, but I can't deny that it does seem to work. Could argue that it even looks a little better than before.

PR #1589, opened 2025-05-23
Revert Track Selection PR 1230 for 17.x

KatrinIhler:
This reverts #1230 from r/17.x as I would rather not introduce such a UI change in a minor release. Will revert the revert for develop.


KatrinIhler:
Gonna merge this because I need this for the next 17.x release of Opencast. We can still discuss whether this was the right decision in the next Dev meeting and, if needed, release the reverted change for 17.

PR #1590, opened 2025-05-23
Re-Introduce Track Selection PR 1230 to develop

KatrinIhler:
The revert of the revert in #1589, re-introducing the new track selection of #1230 to develop. Hopefully doing it this way will mean no problems with future merges between the branches.


KatrinIhler:
Since this basically only reverts to the status quo, I'm gonna merge this without waiting for a review. Feel free to hit me over the head for this later. ;)

PR #1593, opened 2025-06-10
Add Zoom dropdown

Arnei:
Adds a little dropdown next to the zoom slider. The dropdown display the length of the current zoom "snippet". This is supposed to give users a better sense of how long the timeline actually is when zooming. The dropdown also lets users skip to commonly used zoom values (the "commonly used values" are arbitrarly defined by me and will only show if the video is long enough for them to make sense to appear).

Contains #1587.

Screenshot:
Bildschirmfoto vom 2025-06-10 14-24-58

### DIFF
--- a/src/main/CuttingActions.tsx
+++ b/src/main/CuttingActions.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 
-import { BREAKPOINTS, basicButtonStyle, customIconStyle, undisplay } from "../cssStyles";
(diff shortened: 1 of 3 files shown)


snoesberger:
I did some tests with a ~2h recording and the dropdown for the zoom level worked as expected. The dropdown offered the options to zoom in 10s, 1min, 5min or full (All). For my tests I did use the following browsers: Firefox, Safari, Chrome, Edge.

PR #1594, opened 2025-06-10
Test PR

gregorydlogan:
- **Incorrect variable name**
- **Including a link to the release that we're updating to**
- **Branch has been renamed**
- **Detecting branch version, and using appropriate test instance**
- **Using amends to keep from carting around the full build history.**
- **Altering base branch based on which branch the tag comes from**
- **Incorrect variable name**
- **Fixing yaml formatting**
- **Updating the release process**
- **Breaking PR filing out from release generation since that's really two things**
- **Detecting the branch in the case of develop**
- **Switching from hardcoded owners to using repo owners.  In upstream cases this is the same thing, in testing/personal cases it's not!**
- **Cleaning up workflow and step names**
- **Fixing some concurrency issues.  test updates weren't doing concurrency at all leading to potential races, mainline repo just had its group renamed.**
- **Cleaning up the test instances *before* we generate the index**
- **Using working-directory rather that raw cd where possible**
- **Crap to trigger a build**

### DIFF
--- a/.github/get-release-server.sh
+++ b/.github/get-release-server.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+if [ $# -ne 1 ]; then
+  echo "Usage: $0 OC_VERSION"
+  echo " eg: $0 r/16.x -> Returns the current correct server for r/16.x"
+  exit 1
+fi
+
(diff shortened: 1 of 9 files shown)


gregorydlogan:
Wrong repo again

PR #1632, opened 2025-09-09
Revert "fix accidentally publishing of videos"

jduehring:
This reverts commit 98ba1934f08f623f1537a9675cc452d803766a91. I accidentally pushed it directly to the public repo. 

Learned my lesson and won't happen again. Sorry!


Arnei (commented):
Looks like a reasonable fix to me. Since it seems you have learned your lesson, I think we do not need to apply this revert just so you can submit the same commit again.

Issue #105, opened 2021-01-18
Implement an "Undo" Button


Arnei:
Sometimes we have happy little accidents, and in such a case it would be nice if we could undo it with a simple click. An "Undo" button sounds like the solution. As states are managed in redux, implementing undo functionality should be managable.

Things to consider:
* What should the user be able to "undo"? (In the cutting view at least cut, delete and merge, but maybe more?)
* How many steps into the past should the user be able to "undo"?
* Should the functionality be present in other views than cutting and how should it behave?
* Where there is need for Undo, there is need for Redo.

PR #1647, opened 2025-10-29
Chapter Editor

Arnei:
Adds the new menu option "Chapter Editor". Handles avtt file that specify chapters.
Chapters are much like cutting segments in that they are continous, covering the whole video duration.
As such the chapter editor is highly similar to the subtile editor, but incooperates elements from the cutting view as well.

Requires backend changes. Won't achieve anything without backend changes: https://github.com/opencast/opencast/pull/7142

----------

Open points:
- The video player we are using in the editor does not render chapters. Therefore I am not sure what we should do with it in the chapter editor view. We could:
  - Display chapters as subtitles on the video, same as how it works in the subtitle view. This could potentially confuse users though.
  - Don't display anything on the video.
  - Make the video player capable of displaying chapters (likely lots of work, not recommended)
- The text for the chapter segments on the timeline is hard to read thanks to the waveform
- Displaying chapters and cutting segments alongside each other could be potentially confusing?
- Should there be more utility buttons in the list? What should they do?

---------

Quick demo video:
[Bildschirmaufzeichnung vom 2025-10-29 10-15-09.webm](https://github.com/user-attachments/assets/456fb2c9-114a-4209-bb2f-c94b67436f37)

If you want to take a look at this using the test link generated by github actions, append `?chapters.show=true` to the link to make the chapter editor show up.


majosch:
Tested this in a local container environment and on the test deployment linked above.

* Pressing "Enter" in the text-field creates a new chapter, but the "start" time for the new created chapter is after the overall video length.

[Bildschirmaufzeichnung vom 2025-11-05 09-43-43.webm](https://github.com/user-attachments/assets/6caff253-92b9-4e2f-ae2f-d130ce176bbc)

* Doing this multiple times causes incorrect markers on the timeline

[Bildschirmaufzeichnung vom 2025-11-05 10-10-23.webm](https://github.com/user-attachments/assets/3bd8eddd-82e1-49da-8b42-b1312e2f571d)

* Maybe the safest option would be to not create a new chapter if pressing "Enter" in the the text field?

* The readability of the chapter names in the timeline is really bad. I'm not sure if a font-color change would improve this?

* The Preview-Mode-Switch does not seem to have any effect. Deleted segments are always skipped.

[Bildschirmaufzeichnung vom 2025-11-05 10-18-15.webm](https://github.com/user-attachments/assets/f0dbfaf7-891e-4f68-9dce-b9b167b4e140)

* Maybe change icon and wording from scissors and "Cut" to a bookmark icon and "Add"

* Deleting chapters with by merging segments (and possibly overwriting existing chapter titles) seems unnecessary complicated to me. Maybe it would be cleaner to just display an delete button - next to the text fields (subtitle editor) and/or at the action-buttons beneath the timeline? Clicking this button would be functionally equivalent to doing a "Merge Right" on the previous chapter?


From the open points:

> The video player we are using in the editor does not render chapters. Therefore I am not sure what we should do with it in the chapter editor view.

Maybe it would be better to display the chapters at the top of the video?

> The text for the chapter segments on the timeline is hard to read thanks to the waveform
> Displaying chapters and cutting segments alongside each other could be potentially confusing?

I think showing the cutting segments should be kept - but maybe we the chapter segments could move into a second "timeline" (without timelinepreviews)?

> Should there be more utility buttons in the list? What should they do?

Maybe a "Undo"-Button could be implemented?

Arnei:
Thanks for reviewing!

> Maybe it would be better to display the chapters at the top of the video?

Could you illustrate how you would expect that to look like?

> I think showing the cutting segments should be kept - but maybe we the chapter segments could move into a second "timeline" (without timelinepreviews)?

So more similar to the approach in the subtitle view, but having the chapters displayed over the waveform instead of the cutting segments?

> Maybe a "Undo"-Button could be implemented?

That would be nice, not only for the chapter editor but the editor as a whole. As such, I think it would be better to implement Undo functionality in a holistic manner, which is beyond the scope of this PR. (Related open issue: https://github.com/opencast/opencast-editor/issues/105)

geichelberger:
I would change the wording, `Cut` seems misleading, wouldn't something like `Mark`, `Add Mark` or `Add Chapter` fit better?

Arnei:
I changed the wording to "Add" in the most recent commit, as suggest by majosch. I don't know whether "Mark" or "Add Mark" wouldn't be better though.

geichelberger:
Is the langstring add-button not used anywhere?

Arnei:
Thanks for catching that, should be in use now.

majosch:
>> Maybe it would be better to display the chapters at the top of the video?

> Could you illustrate how you would expect that to look like?

I tried but failed because all "advanced" snipping tools (shutter, flameshot) I tried didn't work because lacking Wayland-support :facepalm: 

But I think it's OK to leave it as it is.

>> I think showing the cutting segments should be kept - but maybe we the chapter segments could move into a second "timeline" (without timelinepreviews)?

> So more similar to the approach in the subtitle view, but having the chapters displayed over the waveform instead of the cutting segments?

That was my first idea - but honestly I don't think a change would make it more intuitive... so for me it would be fine to leave it as it is.

* It would be nice if the the chapter title in timeline would use a black font in `Light Mode` - like in `Light (High contrast)`.

grafik

* In `Light (High contrast)` the video view is slightly overlapping the timestamp boxes:

grafik

* When adding a new chapter, the title is copied from the previous chapter - I think it would be better to leave the text box empty.

geichelberger:
The current chapter timeline looks messy. It would be cleaner and easier to read if the waveform were centered in the lower part and the chapter part had less height.

For example, similar to that:
grafik

majosch:
Just re-tested this - all open issues have been resolved! :)

Huge thanks to @Arnei

geichelberger (approved):
LGTM

PR #1650, opened 2025-11-04
Upgrade and adjust hotkeys keyboard shortcuts

ferishili:
This PR consists of the following changes:

- React-hotkeys-hook package upgrade: `4.6` to `5.2`
- Upgrade changes applied
- Zoom in & out key-combo have been changed from `z / t` to `e / r`
- Main focus to the current main menu applied to gain attention of hotkeys
- Some places have now better shortcut behaviors by applying prevent defaults etc.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
         "react-dom": "^18.3.1",
         "react-draggable": "^4.4.6",
         "react-final-form": "^6.5.9",
-        "react-hotkeys-hook": "^4.6.1",
+        "react-hotkeys-hook": "^5.2.1",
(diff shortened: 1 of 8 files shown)


KatrinIhler:
Is there a specific reason this was merged into 17.x?

ferishili:
> Is there a specific reason this was merged into 17.x?

I think @Arnei could answer your question better :D

Arnei:
No idea. Probably because it improves hotkey behaviour?

KatrinIhler:
Unless the prior behavior is a significant problem for people, I'd rather stuff like this goes into stable next time, especially since this changes hot keys.

Issue #1563, opened 2025-03-13
Remove languages that contain no translations


Arnei:
We offer quite a few different languages. However, most of them are just english again (the only ones that have reasonable level of translation are german, greek and spanish). We should not lie to our users about which languages are available.
We could:
- Remove languages from crowdin.
  - At least, all the languages that have been sitting at 0%.
- Remove translations during build that are below a certain threshold.
- Don't download translations in the crowdin workflow if they are below a certain threshold.

oas777:
Could you please also start all remaining languages with a capital letter?

JulianKniephoff:
I don't particularly care **where** it is (in the app, the CI pipeline, etc.), but there should be **some** place that states which languages we support aside from Crowdin. If the mere presence on Crowdin determines the presence in the language switcher, and we don't want (mostly) incomplete translations, we can only ever accept (reasonably) complete translations as contributions. 🤔

PR #1655, opened 2025-12-03
Don't offer languages with little or no translation

Arnei:
Helps with #1563.

The language selector offers quite the range of languages. However, for most of them there are not actual translations, so an eager user might be quite disappointed when they try to change to their favourite language and nothing happens.

This patch reduces the number of languages so that only languages that currently have a translation percentage of 70% or higher are still selectable.

Is there a better way to go about this than hardcoding which languages are available? Probably. But this works for now.


Arnei:
Update: All languages with no translation have now been removed from crowdin. So this PR effectively only removes Chinese Traditional (36%), Dutch (35%) and Chinese Simplified (1%).

src/i18n/lngs-generated.ts
```diff
@@ -1,20 +1,7 @@
 export const languages = new Map<string, string>([
-  ["am", "am-ET"],
-  ["cs", "cs-CZ"],
-  ["de-AT", "de-AT"],
   ["de-DE", "de-DE"],
```
mtneug, line 2: Should this be just the following?

```
  ["de", "de-DE"],
```

Issue #1391, opened 2024-06-27
Button responsiveness in different views


snoesberger:
If you are on a small screen, some of the buttons disappear from the view:
![grafik](https://github.com/opencast/opencast-editor/assets/103581881/7db59d42-ca15-492f-b917-7e54131386b8)
<img width="969" alt="grafik" src="https://github.com/opencast/opencast-editor/assets/103581881/77b896e4-d252-4ef2-97da-95fa40fd1405">
<img width="954" alt="grafik" src="https://github.com/opencast/opencast-editor/assets/103581881/3da81baf-4bed-440d-bc2a-cce23166fc36">

The responsiveness of the buttons should be improved.

Arnei:
At least the cutting view should be improved by https://github.com/opencast/opencast-editor/pull/1160

Arnei:
This should be improved now.

Issue #928, opened 2022-12-06
Subtitle editor, overlapping UI elements


cgruber-uos:
The UI text "Vorschaumodus" is overlapped by the subtitle add / delete function. And the milliseconds of the total playing time is cropped by the browsers size: 

![2022-12-06 13_42_22-Greenshot](https://user-images.githubusercontent.com/101632236/205916196-15afab10-6789-4599-bbac-1c0732fd2467.png)

![2022-12-06 13_42_28-Greenshot](https://user-images.githubusercontent.com/101632236/205916216-bdf7b5e1-5abb-4dd0-88ef-b2429bfafb9c.png)

Special for my resolution (1440 px width) or a general problem?

cgruber-uos:
fixed by https://github.com/opencast/opencast-editor/pull/908

Arnei:
This has been improved with #1664 and other PRs.

PR #1664, opened 2026-01-28
Help with Subtitle Editor responsiveness

Arnei:
Follow up on #1160. Fixes the rest of #1391 and #928.

Should help with the video controls and the buttons on the top right overflowing out of sight for smaller window widths in the subtitle editor view. They will eventually still disappear if the width gets small enough, but  the subtitle editor is arguably not designed to work for such small widths anyway.

Bildschirmfoto vom 2026-01-28 12-38-30

### How to test this

Open the subtitle editor (and every other view that has video controls) and test out different screen sizes. (Views can be enabled in the config or via query parameter, e.g. `subtitles.show=true`)

### DIFF
--- a/src/cssStyles.tsx
+++ b/src/cssStyles.tsx
@@ -426,3 +426,9 @@ export const undisplay = (maxWidth: number) => css({
     display: "none",
   },
 });
+
+export const undisplayContainer = (maxWidth: number) => css({
+  [`@container (max-width: ${maxWidth}px)`]: {
+    display: "none",
+  },
+});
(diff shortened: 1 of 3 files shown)


gregorydlogan:
This looks good to me, and I'd be willing to merge this in, but I'm not sure if what I'm seeing here is behaviour caused by this PR or not.  Watch the Delete, Upload, and Download buttons:

https://github.com/user-attachments/assets/814ffa93-f3f0-4673-aa58-0f95cf8f51f2

There's clearly some fighting in the logic around whether to display them with text, or without.  That being said, aside from semi-randomly showing text they stay where I would expect them to and do *work*.

Arnei:
The behaviour around the Delete, Upload and Download buttons is caused by this PR and intentional. They now flex-wrap instead of overflowing out of the window, and the text is removed when the window width goes below a certain magic number. Nothing random about it.

I think ideally, we would remove the text from the buttons when they are just about to wrap around, and then only wrap around if necessary. However, that requires far more careful javascripting than I am willing to invest here.

gregorydlogan (approved):

PR #1672, opened 2026-01-30
Add new config "stopOnTyping" for subtitles

### LINKED ISSUE #1397: Subtitle Editor: Option "pause while typing"
It would be nice to have a "pause while typing" option for the subtitle editor. If the option is enabled and the user starts typing, the video would pause.

Arnei:
Fixes #1397.

Adds a new option to have video playback in the subtitle editor pause automatically if the user starts typing in one of the subtitle boxes.

### DIFF
--- a/editor-settings.toml
+++ b/editor-settings.toml
@@ -111,6 +111,11 @@
 # Default: "captions"
 #mainFlavor = "captions"
 
+# If set to true, video playback stops when typing in the subtitle fields
+# Type: boolean
+# Default: false
+#stopOnTyping = false
+
 [subtitles.languages]
 ## A list of languages for which new subtitles can be created
 # For each language, various tags can be specified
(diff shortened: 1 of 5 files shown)


editor-settings.toml
```diff
@@ -111,6 +111,11 @@
 # Default: "captions"
 #mainFlavor = "captions"
 
+# If set to true, video playback stops when typing in the subtitle fields
```
gregorydlogan, line 114: This needs a change in `opencast/opencast` right?  I don't see a PR for it, unless I've missed it?

Arnei: Good catch, I keep forgetting that changes I make to the toml here do not automatically propagate to the toml in opencast/etc/ui-config/editor ^^'

Arnei: There you go https://github.com/opencast/opencast/pull/7575

PR #1675, opened 2026-02-02
Allow typing time in current time field

### LINKED ISSUE #67: Turn the time display into an input field
To allow the user precise cutting beyond the capabilities of the scrubber, it would be nice to allow the user to enter the exact time they want to place a cut. One way to achieve this without adding any new UI elements could be to turn the time display next to the play button into an input field.

Arnei:
Fix #67

This patch lets users change the current time by typing. They can do so by clicking on the current time which will then turn into an input field (focusing and pressing enter also works). They can then change the time by typing.

The changed time is only applied on blur or pressing "Enter". Any inputs that do not fit the syntax are ignored.

Bildschirmfoto vom 2026-02-02 11-13-47

### DIFF
--- a/src/main/Cutting.tsx
+++ b/src/main/Cutting.tsx
@@ -135,6 +135,7 @@ const Cutting: React.FC = () => {
           selectIsMuted={selectIsMuted}
           selectVolume={selectVolume}
           selectIsPlayPreview={selectIsPlayPreview}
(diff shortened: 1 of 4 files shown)


ferishili (approved):
Thanks @Arnei!
This works fine as described!

PR #1677, opened 2026-02-04
Add configurable hotkeys

Arnei:
Allows users to configure their own hotkeys.
This should empower users to choose the hotkeys that work for them, instead of having to use the cumbersome "Shift + Alt + something" combinations.

Configured hotkeys are persisted in local storage. To that end, this patch introduces the redux-persist package and all the necessary boilerplate that it requires.

This also makes more changes to KeyboardControls.tsx than strictly necessary in the hopes of making it a little less complicated.

Bildschirmfoto vom 2026-02-04 12-52-50

Bildschirmfoto vom 2026-02-04 12-52-38

### DIFF
--- a/src/main/KeyboardControls.tsx
+++ b/src/main/KeyboardControls.tsx
@@ -1,21 +1,28 @@
+import React from "react";
 import { css } from "@emotion/react";
 import { ParseKeys } from "i18next";
-
-import React from "react";
-
 import { useTranslation, Trans } from "react-i18next";
-import { getGroupName, KEYMAP, rewriteKeys } from "../globalKeys";
-import { useTheme } from "../themes";
-import { titleStyle, titleStyleBold } from "../cssStyles";
+import { getGroupName, IKey, IKeyGroup, rewriteKeys } from "../globalKeys";
+import { Theme, useTheme } from "../themes";
+import { basicButtonStyle, titleStyle, titleStyleBold } from "../cssStyles";
+import { useRecordHotkeys } from "react-hotkeys-hook";
+import { useAppDispatch, useAppSelector } from "../redux/store";
+import { selectKeymap, setHotkey } from "../redux/hotkeySlice";
+import { Modal, ModalHandle, ProtoButton } from "@opencast/appkit";
+import { LuPen } from "react-icons/lu";
 
-const Group: React.FC<{ name: ParseKeys, entries: { [key: string]: string[][]; }; }> = ({ name, entries }) => {
+const Group: React.FC<{
+  id: string
+  entries: IKeyGroup
+  openEditModal: (group: string, action: string, actionTitle: string) => void
+}> = ({ id, entries, openEditModal }) => {
 
   const { t } = useTranslation();
   const theme = useTheme();
 
   const groupStyle = css({
     display: "flex",
-    flexDirection: "column" as const,
+    flexDirection: "column",
     width: "460px",
     maxWidth: "50vw",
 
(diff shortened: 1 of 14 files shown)


src/main/KeyboardControls.tsx
```diff
@@ -163,9 +236,104 @@ const KeyboardControls: React.FC = () => {
...
+  const dispatch = useAppDispatch();
+  const { t } = useTranslation();
+  const theme = useTheme();
+
+  const setNewKeys = () => {
+    stop();
+
+    dispatch(setHotkey({
+      group: group, // "timeline"
+      action: action, // "left"
```
ferishili: I think these comment messages were unintentionally left in. Can we remove them?

src/main/KeyboardControls.tsx
```diff
@@ -163,9 +236,104 @@ const KeyboardControls: React.FC = () => {
...
+      key: Array.from(keys).join(" + "),
```
ferishili, line 326: What happens if someone has a keyboard layout where "+" is there and has been used in a key combination with other keys?

ferishili (changes requested):
Thanks @Arnei,

Here are a few points I'd like to highlight after testing the current behavior:

1. **Saving without a key combination clears the entry silently**

   * Right now, saving without entering a key combination results in an empty value being saved.
   * At minimum, the UI should notify the user that the existing shortcut is being removed.
   * Ideally, saving an empty value should be prevented entirely, and instead a dedicated **Delete** option should be provided (either inside the modal or, even better, next to the edit button with a confirmation step).

2. **Empty key combinations are displayed as a small gray square**

   * When an entry has no assigned key combination, the UI currently shows a small gray square.
   * It would be clearer to display a label such as **"Not defined"** or **"No combination"** instead.

3. **Duplicate key combinations are currently allowed**

   * At the moment, it's possible to assign the same shortcut (e.g., `ArrowLeft`) to multiple entries.
   * The system should detect duplicates and notify the user.
   * Ideally, the UI should offer the option to replace the existing assignment with confirmation.

4. **`resetHotkey` and `resetAllHotkeys` reducers delete instead of reset**

   * The reducers `resetHotkey` and `resetAllHotkeys` in `src/redux/hotkeySlice.ts` currently remove values instead of restoring defaults, and they are not being used.
   * It would be better if these reducers restored the default key combinations instead.
   * I'd also suggest adding **"Reset to default"** (per entry) and **"Reset all to default"** actions on the keyboard controls page.

src/main/KeyboardControls.tsx
Arnei: You mean if somebody wants to use the "+" key in their hotkey sequence? That would break the hotkey, as "+" is the default sequence split key.
Tbh I was kind of hoping that I could ignore this, as I do not expect many user will want to use the "+" as their hotkey, and "+" is also already used in the ui for delimiting hotkey sequences which should deter users further from trying to use "+" as a hotkey.

ferishili: > You mean if somebody wants to use the "+" key in their hotkey sequence? That would break the hotkey, as "+" is the default sequence split key.

Yes!

> Tbh I was kind of hoping that I could ignore this, as I do not expect many user will want to use the "+" as their hotkey, and "+" is also already used in the ui for delimiting hotkey sequences which should deter users further from trying to use "+" as a hotkey.

You would never know! :D
And I would suggest to handle this now, instead of bug report later!

Arnei:
Thanks for the nice suggestions. I hope I managed to address them reasonably well.

ferishili:
Thanks @Arne,
the requested changes are covered now!

However, there are still a couple of small things left to address:

1. The **space** key is shown as an empty box after saving, and it is not displayed when entering it before saving.

Screenshot 2026-04-16 at 14 34 03

2. The key combination is currently **not limited** (when spaces are used between keys?). This makes it possible to enter a lot of keys in a single shortcut, which should not be allowed.
   Please limit the combination to a maximum of **3 keys** in any case.

Screenshot 2026-04-16 at 14 39 52

ferishili (approved):
@Arnei thanks!
I can confirm that everything works as expected!

Issue #1184, opened 2023-10-23
Feature Request: Make comments available


Rillke:
The old editor had a comments tab. It would be nice, if one could edit comments in the new editor, too.

![image](https://github.com/opencast/opencast-editor/assets/2311611/d77e3d54-6a85-409c-965b-a70c9c94cb65)

Rillke:
Appearently, no one misses/requires this feature.

mtneug:
I think it would be nice to have this. Was there a poll and nobody wanted this? IMO being able to comment is an underrated feature of Opencast.

Rillke:
> I think it would be nice to have this. Was there a poll and nobody wanted this? IMO being able to comment is an underrated feature of Opencast.

@mtneug  I didn't attend the workshop @oas777 can tell you more. IIRC it was a combination of nobody really wanted it and Olaf was in fear of too many tools/ sidebar clutter.

That said, our Opencast substitute staff also wasn't happy dropping this. Please feel free to reopen but we probably won't fund it.

oas777:
We're not interested in this (anymore) because comments are being used in the admin UI only. If @Rillke "probably won't fund it", this is a "won't fix" because no one else mentioned this in Jena.

mtneug:
Please leave this open. We are interested in this.

mtneug:
A student now works on this issue for us.

PR #1685, opened 2026-03-03
feat: comments

Bennit99:
Adds a comment system for video editing like in the old admin UI.

- Comments.tsx: Renders a scrollable list of comments with nested replies, timestamps, and status badges. Includes forms to add comments (with a required reason dropdown), reply, delete, and toggle resolved status.
- commentSlice.ts: Manages local state for adding, deleting, and updating comments. Tracks uncommitted changes hasChanges and handles temporary local IDs.
- Also configure a temporary vite dev proxy to bypass CORS for testing with loacal opencast instance 

- For testing create and delete some comments and replies

[Issue: Feature Request: Make comments available](https://github.com/opencast/opencast-editor/issues/1184)

[backend PR](https://github.com/opencast/opencast/pull/7452)

### DIFF
--- a/public/editor-settings.toml
+++ b/public/editor-settings.toml
@@ -18,7 +18,9 @@ callbackSystem = "OPENCAST"
 
 [opencast]
 # Connect to develop.opencast.org and use the default demo user
-url = 'https://develop.opencast.org'
+# url = 'https://develop.opencast.org'
+# When using the vite proxy, leave url empty so requests go through the dev server
+url = ''
 name = "admin"
 password = "opencast"
 
--- a/public/editor-settings.toml
+++ b/public/editor-settings.toml
@@ -48,3 +48,6 @@ spanish = { lang = "es" }
 
 [thumbnail]
 show = true
+
+[comments]
+show = true
--- a/public/editor-settings.toml
+++ b/public/editor-settings.toml
@@ -18,9 +18,7 @@ callbackSystem = "OPENCAST"
 
 [opencast]
 # Connect to develop.opencast.org and use the default demo user
-# url = 'https://develop.opencast.org'
-# When using the vite proxy, leave url empty so requests go through the dev server
-url = ''
+url = 'https://develop.opencast.org'
 name = "admin"
 password = "opencast"
 
--- a/public/editor-settings.toml
+++ b/public/editor-settings.toml
@@ -50,4 +50,4 @@ spanish = { lang = "es" }
 show = true
 
 [comments]
-show = true
+show = false
--- a/src/i18n/locales/en-US.json
+++ b/src/i18n/locales/en-US.json
@@ -7,6 +7,7 @@
       "chapters-button": "Chapters",
       "thumbnail-button": "Thumbnail",
       "metadata-button": "Metadata",
+      "comments-button": "Comments",
       "keyboard-controls-button": "Keyboard Controls",
       "tooltip-aria": "Main Navigation"
     },
@@ -246,7 +247,8 @@
       "error-text": "An error has occurred. Please wait a bit and try again.",
       "goBack-button": "No, take me back",
       "callback-button-system": "Back to {{system}}",
-      "callback-button-generic": "Back to previous system"
+      "callback-button-generic": "Back to previous system",
+      "loading": "Loading..."
     },
 
     "trackSelection": {
@@ -329,6 +331,46 @@
 
     "chapters": {
       "editTitle": "Chapter Editor"
+    },
+
+    "comments": {
+      "no-comments": "No comments yet. Be the first to comment!",
+      "resolved": "Resolved",
+      "open": "Open",
+      "reply": "Reply",
+      "delete": "Delete",
+      "add-comment": "Add Comment",
+      "placeholder": "Write your comment...",
+      "select-reason": "Select a reason...",
+      "submit": "Submit",
+      "saving": "Saving...",
+      "reply-to": "Reply to",
+      "reply-placeholder": "Write your reply to",
+      "mark-resolved": "Mark as resolved",
+      "cancel": "Cancel",
+      "pending": "Pending",
+      "pending-tooltip": "Pending changes need to be saved in the Finish tab.",
+      "reasons": {
+        "cutting": "Cutting required",
+        "review": "Review required",
+        "async": "A/V tracks asynchronous",
+        "audio_issue": "Missing or defective audio track",
+        "cancelled": "Canceled event",
+        "conflicting_metadata": "Conflicting metadata",
+        "improper_point": "Improper in or out point",
+        "missing_agreement": "Missing agreement",
+        "other": "Other",
+        "privacy": "Privacy concern",
+        "segmentation": "Inaccurate segmentation",
+        "unknown_creator": "Creator unknown",
+        "video_issue": "Video distorted or cropped",
+        "wrong_input_format": "Input file format not supported",
+        "wrong_metadata": "Metadata needs correction",
+        "wrong_series_publication": "Wrong series or publication channel",
+        "wrong_workflow": "Wrong workflow",
+        "processing_failure": "Processing failure",
+        "admin_ui_notes": "Notes in Admin UI"
+      }
     },
 
     "keyboardControls": {
--- a/src/main/Comments.tsx
+++ b/src/main/Comments.tsx
@@ -1,0 +1,543 @@
+import { useState } from "react";
+import { css } from "@emotion/react";
+import { useTheme } from "../themes";
+import Select from "react-select";
+import {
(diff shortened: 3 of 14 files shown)


gregorydlogan:
For those curious, I have confirmed that this is a student at Munster.

public/editor-settings.toml
```diff
@@ -48,3 +50,6 @@ spanish = { lang = "es" }
 
 [thumbnail]
 show = true
+
+[comments]
+show = true
```
mtneug: It was requested to disable this feature by default in the interest of keeping the number of elements in the menu small.

```suggestion
show = false
```

vite.config.ts
```diff
@@ -12,6 +12,51 @@ export default defineConfig(() => {
     base: process.env.PUBLIC_URL || "",
     server: {
       open: true,
+      proxy: {
```
mtneug: Needs to be reverted before merge?

public/editor-settings.toml
```diff
@@ -18,7 +18,9 @@ callbackSystem = "OPENCAST"
 
 [opencast]
 # Connect to develop.opencast.org and use the default demo user
-url = 'https://develop.opencast.org'
+# url = 'https://develop.opencast.org'
+# When using the vite proxy, leave url empty so requests go through the dev server
+url = ''
```
mtneug: Needs to be reverted before merge.

src/main/MainMenu.tsx
```diff
@@ -88,6 +95,12 @@ const MainMenu: React.FC = () => {
         bottomText={t(MainMenuStateNames.thumbnail)}
         ariaLabelText={t(MainMenuStateNames.thumbnail)}
       />}
+      {settings.comments?.show && <MainMenuButton
```
mtneug: Is the `?.` in `comments?.` necessary?

mtneug (changes requested):
Works as expected. Only some small comments from my side. Since I don't really work on the front-end, it would be great to also get a review from someone more familiar with the codebase.

src/main/MainMenu.tsx
Bennit99: I added it for backwards compatibility, but as the settings are merged with the default settings anyway, its not really necessary.

public/editor-settings.toml
Bennit99: Done

vite.config.ts
Bennit99: Done

public/editor-settings.toml
```diff
@@ -48,3 +48,6 @@ spanish = { lang = "es" }
 
 [thumbnail]
 show = true
+
+[comments]
+show = false
```
gregorydlogan, line 53: This should *probably* be true, since this config is what's backing `editor.opencast.org` and we want to show off all of the bells and whistles.

gregorydlogan (commented):
I'm not confident enough to review this properly.  @Arnei can you, or delegate to someone who *would* be confident reviewing this?

src/i18n/locales/en-US.json
```diff
@@ -331,6 +333,46 @@
       "editTitle": "Chapter Editor"
     },
 
+    "comments": {
+      "no-comments": "No comments yet. Be the first to comment!",
```
Arnei, line 337: Would make sense for a social media post or similar, but there is no reason to incentivize posting comments in Opencast. I think "No comments yet." is enough.

src/main/Comments.tsx
```diff
@@ -0,0 +1,543 @@
...
+  const [commentReplyIsResolved, setCommentReplyIsResolved] = useState(false);
+
+  const [newCommentText, setNewCommentText] = useState("");
+  const [commentReason, setCommentReason] = useState("");
+
+  // Comments are fetched together with video info via fetchVideoInformation
+
+  // Handlers
+  const handleSaveComment = () => {
+    if (!settings.id || !newCommentText || !commentReason) {
```
Arnei, line 56: Not like it is causing problems, but why are we checking for `settings.id` in the save handlers?

src/main/Comments.tsx
```diff
@@ -0,0 +1,543 @@
...
+    maxWidth: "100%", // Prevent children from overflowing
+    // Modern scrollbar styling
+    "&::-webkit-scrollbar": {
+      width: "8px",
+      background: theme.background,
+    },
+    "&::-webkit-scrollbar-thumb": {
+      background: theme.text,
+      borderRadius: "5px",
+    },
```
Arnei, line 143: This is not bad or anything, but definitely different from every other scrollbar in the editor. I don't think the stylistic inconsistency is worth it?

src/main/Comments.tsx
```diff
@@ -0,0 +1,543 @@
...
+    return (
+      <div css={containerStyle}>
+        <div css={loadingStyle}>{t("various.loading")}</div>
+      </div>
+    );
+  }
+
+  return (
+    <div css={containerStyle}>
+      <h2 css={titleStyleBold(theme)}>{t("mainMenu.comments-button")}</h2>
```
Arnei, line 350: For consistency with other views, could you center the title?

src/main/Comments.tsx
```diff
@@ -0,0 +1,543 @@
...
+                        </Badge>
+                      )}
+                    </div>
+                    <span css={css({
+                      display: "flex",
+                      alignItems: "center",
+                      gap: "6px",
+                      fontSize: "0.9em",
+                      color: theme.text,
+                      opacity: 0.7,
```
Arnei, line 427: Having a fixed opacity can lead to low contrast in the high-contrast color modes. You'll likely either want to make your opacity level a theme variable, or adjust your text color.

src/main/Comments.tsx
```diff
@@ -0,0 +1,543 @@
...
+          />
+
+          {/* flex row container for select and buttons */}
+          <div css={rowContainerStyle}>
+            {/* "Select a reason" dropdown */}
+            <div css={selectContainerStyle}>
+              <Select<{ value: string; label: string }>
+                styles={selectFieldStyle(theme)}
+                value={commentReason
+                  ? { value: commentReason, label: t(`comments.reasons.${commentReason}` as never) }
```
Arnei, line 471: The more semantically correct type here would be `ParseKeys` from i18next. Might also make sense to type commentReason in a way that no typecast is needed? (same for the other two occurences of this)

src/i18n/locales/en-US.json
```diff
@@ -331,6 +333,46 @@
...
+        "privacy": "Privacy concern",
+        "segmentation": "Inaccurate segmentation",
+        "unknown_creator": "Creator unknown",
+        "video_issue": "Video distorted or cropped",
+        "wrong_input_format": "Input file format not supported",
+        "wrong_metadata": "Metadata needs correction",
+        "wrong_series_publication": "Wrong series or publication channel",
+        "wrong_workflow": "Wrong workflow",
+        "processing_failure": "Processing failure",
+        "admin_ui_notes": "Notes in Admin UI"
```
Arnei, line 372: I *think* this should be `adminui_notes`?

src/redux/commentSlice.ts
```diff
@@ -0,0 +1,212 @@
...
+     */
+    addComment: (state, action: PayloadAction<{
+      reason: string;
+      text: string;
+    }>) => {
+      const newComment: Comment = {
+        id: nanoid(),
+        creationDate: new Date().toISOString(),
+        author: "",  // Backend stamps the actual author via SecurityService
+        displayName: "",
```
Arnei, line 55: Any chance we could have the display name of the current user here? It is a bit disorienting that new comments do not have the name of a user associated with them in the ui.

Arnei (changes requested):
Overall this looks very good to me! The stuff below is all pretty nitpicky.

I think I might have liked it better if the <Comments> component was not quite as monolithic and split into more components, but I can see why you did that and it should not cause any trouble.

One thing I would like to see changed is the use of colors, namely I think there should be less of them. The color scheme of the editor is decidedly grayscale. Colors should not be used to make the ui look nice, but to explicitly draw the users attention to certain elements. For example, red signifies danger, so having the delete buttons in red makes sense to me. But the Submit/Reply/Cancel buttons at the bottom right do not require any color in my opinion. I'd say the same goes for the badges. Here there is also the additional issue that their color contrast to the background is quite low in light mode, making them hard to read.

mtneug:
FYI: I will take over the development of this PR as @Bennit99 sadly no longer works as a student for us. It might take some time, but I will work on the mentioned issues.

public/editor-settings.toml
mtneug: OK, then I was mistaken to request the opposite.

PR #1694, opened 2026-04-08
Fix focus loss caused by sliders affecting keyboard shortcuts (follow-up to #1650)

ferishili:
This PR is a follow-up/fix for the new keyboard shortcuts introduced in #1650.

The issue occurred because the sliders (zoom and volume) were taking focus during interaction and not returning it afterward to the main element. As a result, keyboard shortcuts stopped working once users adjusted those sliders.

**Solution**

This PR introduces a `MainFocusContext` that allows components to restore focus back to the main workspace after slider interaction completes.

Specifically:

* A focus context is added and provided at the main content level
* Components using sliders consume this context
* Slider `onChangeCommitted` events trigger restoring focus to the main workspace
* Keyboard shortcuts remain functional after slider usage
* The `MainFocusContext` can be used also in other places later on!

**Result**

After adjusting zoom or volume sliders:

* focus returns to the main element automatically
* keyboard shortcuts continue working as expected
* behavior stays consistent with the intended shortcut workflow

## UPDATE

This PR had a revert on the above mentioned solution and now got newer changes: https://github.com/opencast/editor/pull/1694#issuecomment-4223842080

### DIFF
--- a/src/globalKeys.ts
+++ b/src/globalKeys.ts
@@ -48,97 +48,161 @@ export interface IKeyGroup {
 export interface IKey {
   name: string;
   key: string;
+  options: object,
   splitKey?: string;
 }
+
+const hotkeysDefaultOptions: object = {
+  preventDefault: true,
+  enableOnFormTags: ["slider"],
+};
+
+export const subtitleListHotkeysDefaultOptions: object = {
+  preventDefault: true,
+  enableOnFormTags: ["input", "select", "textarea", "slider"],
+};
+
+const cuttingZoomInSplitKey = ";";
 
 export const KEYMAP: IKeyMap = {
   videoPlayer: {
     play: {
       name: "keyboardControls.videoPlayButton",
       key: "Shift+Alt+Space, Space",
+      options: {
+        ...hotkeysDefaultOptions,
+      },
     },
     previous: {
       name: "video.previousButton",
       key: "Shift+Alt+Left",
+      options: {
+        ...hotkeysDefaultOptions,
+      },
     },
     next: {
       name: "video.nextButton",
       key: "Shift+Alt+Right",
+      options: {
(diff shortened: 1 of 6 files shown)


Arnei:
I can reproduce the issue you've described and this PR does fix it. However, I am not quite happy with the solution.
- This solution moves the focus to the start of the view. If you are navigating by keyboard, this sudden jump in focus is very disorienting.
- This breaks keyboard controls on the slider. For example, I could adjust the slider by pressing "<-" or "->" on my keyboard. Now, the slider loses focus after a single keystroke.

Do you think there might be another solution? Since the issue seems to lie with key events being caught by the sliders, maybe the sliders could let the key events they don't need propagate? E.g. keep behaviour for "<-" and "->", but let "Space" propagate to trigger the video play/pause?

ferishili:
@Arnei, thanks for the review and for catching that!

I've updated the implementation to use the `enableOnFormTags` option provided by `react-hotkeys-hook`. To make this work precisely, I've made the following changes:

* **Role Injection:** I explicitly added `role="slider"` to the Sliders in both locations. This allows the hotkey logic to identify them specifically.
* **Centralized Configuration:** To maintain reusability and keep the code cleaner, I've bundled these options into the `KEYMAP...` constant. Now, any call to `useHotkeys` can pull the correct options from a single source of truth.

Please take another look. Since this touched the configuration for all hotkeys, a quick sanity test on the general hotkey behavior everywhere would be much appreciated!

Arnei (approved):
This works and looks reasonable to me, thank you!

I've noticed some weird behaviour, where if you drag the zoom slider with the mouse, then click on the timeline, the "-->" and "<--" keys will move both the zoom slider and the scrubber. But that behaviour was already present previously, so it should not block this PR.

PR #1700, opened 2026-05-06
Ensure node version matches parent pom

gregorydlogan:
This PR checks to see that *our* node vesrion matches hte upstream parent's pom's version.

This has hte side effect of updating the current versino of node in use, especially once https://github.com/opencast/opencast/pull/7636 is merged.


gregorydlogan:
Note that this should be failing now, since the changes it makes depend on https://github.com/opencast/opencast/pull/7636 being merged first.

PR #1705, opened 2026-05-19
Update to React 19

### LINKED ISSUE #1643: Update to React 19; blockers
Appkit finally supports React 19, but there are yet more libraries that do not. In particular, `react-beforeunload` and `use-resize-observer` still require updates.

`react-beforeunload` is fortunately small enough that we can just copy the code into our own codebase or write it ourselves if we want to.

`use-resize-observer` is a bit bigger and requires additional changes beyond bumping the version number. I'm inclined to wait until the project developer finds some spare time, since they are obviously still invested in the project.

Arnei:
Fixes #1643.

`react-beforeunload` added support for React 19, and `use-resize-observer` was replaced.

Bunch of other dependencies received major version updates too.


src/main/SubtitleTimeline.tsx
```diff
@@ -46,10 +46,10 @@ const SubtitleTimeline: React.FC = () => {
   const subtitleId = useAppSelector(selectSelectedSubtitleId, shallowEqual);
 
   const ref = useRef<HTMLDivElement>(null);
-  const { width = 1 } = useResizeObserver({ ref });
+  const { width = 1 } = useResizeObserver<HTMLDivElement>({ ref: ref as React.RefObject<HTMLDivElement> });
   const refTop = useRef<HTMLElement>(null);
   const refMini = useRef<HTMLDivElement>(null);
-  const { width: widthMiniTimeline = 1 } = useResizeObserver({ ref: refMini });
+  const { width: widthMiniTimeline = 1 } = useResizeObserver({ ref: refMini as React.RefObject<HTMLDivElement> });
```
owi92, line 52: Urgh I feel like the necessity of this casting here is kind of unfortunate. And weird, because the docs of this hook show an example without that. But I can't seem to get it working without casting either. Oh well...

src/main/VideoPlayers.tsx
```diff
@@ -198,36 +197,36 @@ export const VideoPlayer = React.forwardRef<VideoPlayerForwardRef, VideoPlayerPr
...
       }
     };
 
     // Tries to get video dimensions from the HTML5 elements until they are not 0,
     // then updates the store
     async function updateAspectRatio() {
-      if (ref.current && ref.current.getInternalPlayer()) {
-        let w = (ref.current.getInternalPlayer() as HTMLVideoElement).videoWidth;
-        let h = (ref.current.getInternalPlayer() as HTMLVideoElement).videoHeight;
+      if (ref.current && ref.current) {
```
owi92:
```suggestion
      if (ref.current) {
```

src/main/VideoPlayers.tsx
```diff
@@ -198,36 +197,36 @@ export const VideoPlayer = React.forwardRef<VideoPlayerForwardRef, VideoPlayerPr
...
+        let w = (ref.current as HTMLVideoElement).videoWidth;
+        let h = (ref.current as HTMLVideoElement).videoHeight;
         while (w === 0 || h === 0) {
           await sleep(100);
-          w = (ref.current.getInternalPlayer() as HTMLVideoElement).videoWidth;
-          h = (ref.current.getInternalPlayer() as HTMLVideoElement).videoHeight;
+          w = (ref.current as HTMLVideoElement).videoWidth;
+          h = (ref.current as HTMLVideoElement).videoHeight;
```
owi92: these casts seem to be unnecessary (you didn't add them here, I just noticed)
```suggestion
        let w = ref.current.videoWidth;
        let h = ref.current.videoHeight;
        while (w === 0 || h === 0) {
          await sleep(100);
          w = ref.current.videoWidth;
          h = ref.current.videoHeight;
        }
```

src/main/VideoPlayers.tsx
```diff
@@ -377,7 +376,7 @@ export const VideoPlayer = React.forwardRef<VideoPlayerForwardRef, VideoPlayerPr
       // Renders the current frame in the video element to a canvas
       // Returns the data url
       captureVideo() {
-        const video = ref.current?.getInternalPlayer() as HTMLVideoElement;
+        const video = ref.current as HTMLVideoElement;
```
owi92:
```suggestion
        if (!ref.current) {
          return;
        }
        const video = ref.current;
```

src/main/VideoPlayers.tsx
```diff
@@ -388,7 +387,7 @@ export const VideoPlayer = React.forwardRef<VideoPlayerForwardRef, VideoPlayerPr
         }
       },
       getWidth() {
-        return (ref.current?.getInternalPlayer() as HTMLVideoElement).clientWidth;
+        return (ref.current as HTMLVideoElement).clientWidth;
```
owi92:
```suggestion
        return ref.current?.clientWidth ?? 0;
```

src/main/VideoPlayers.tsx
```diff
@@ -430,23 +429,26 @@ export const VideoPlayer = React.forwardRef<VideoPlayerForwardRef, VideoPlayerPr
       if (!errorState) {
         return (
           <div css={videoPlayerWrapperStyles}>
-            <ReactPlayer url={url}
+            <ReactPlayer src={url}
+              wrapper={"div"}
```
owi92:
```suggestion
        <ReactPlayer
              src={url}
              wrapper={"div"}
```

src/main/VideoPlayers.tsx
```diff
@@ -430,23 +429,26 @@ export const VideoPlayer = React.forwardRef<VideoPlayerForwardRef, VideoPlayerPr
...
               onPlay={onPlay}
               onEnded={onEndedCallback}
               onError={onErrorCallback}
-              tabIndex={-1}
+              // tabindex is currently not properly applied to neither wrapper nor video element
+              tabIndex={-1} // Skip player when navigating page with keyboard
+              crossOrigin="anonymous" // allow thumbnail generation
+              disablePictureInPicture={true}
+              // @ts-expect-error: The Config type does match what we need
               config={playerConfig}
```
owi92, line 455: Maybe just remove the config prop.

src/main/SubtitleListEditor.tsx
```diff
@@ -125,28 +125,18 @@ const SubtitleListEditor: React.FC<{
   }, [addCueAtIndex, defaultSegmentLength, dispatch, subtitle?.cues, subtitleId]);
 
   const listStyle = css({
-    display: "flex",
-    flexDirection: "column",
-    height: "100%",
-    width: "60%",
-    gap: "20px",
+    flex: "1 1 auto",
```
owi92: I think that's the same as 
```suggestion
    flex: "auto",
```

src/main/SubtitleListEditor.tsx
```diff
@@ -188,29 +178,21 @@ const SubtitleListEditor: React.FC<{
...
-            itemData={itemData}
-            itemSize={_index => segmentHeight}
-            // @ts-expect-error: Type is not properly inferred for some reason
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access
-            itemKey={(index, data) => data.items[index].idInternal}
-            width={width ? width : 0}
+      <AutoSizer renderProp={
+        ({ height = 600, width = 800 }) => {
+          return <List
+            style={{ height: height, width: width }}
```
owi92:
```suggestion
            style={{ height, width }}
```

src/main/SubtitleListEditor.tsx
```diff
@@ -188,29 +178,21 @@ const SubtitleListEditor: React.FC<{
...
+            rowComponent={RowComponent}
+            rowHeight={segmentHeight ? segmentHeight : 0}
```
owi92:
```suggestion
            rowHeight={segmentHeight ?? 0}
```

owi92 (approved):
I have tested this and couldn't find any obvious issues. However I am not an avid editor user, so take this with a grail of salt. For my tests I simply tried all functions (basically just clicked through the interface) and compared with https://editor.opencast.org/19.x/ (https://editor.opencast.org/develop/ appears to be broken).

Code looks generally fine too. My comments are just some code suggestions that do equivalent things compared to what's already in place, but they are a little nicer imo.

src/main/VideoPlayers.tsx
Arnei: Afaik we still need that prop, so I can't just remove it.

src/main/SubtitleListEditor.tsx
Arnei: Huh, so it is.

src/main/VideoPlayers.tsx
owi92: I don't think that is the case. The prop is optional and what is being passed here has zero overlap with the type that is required.

Arnei: Which I am blaming on the typing of ReactPlayer being wrong. Afaik these still work, now it is just undocumented.

owi92: Hm alright, I don't have a lot of experience with ReactPlayer and I trust you to make the right call. In any case keeping it in doesn't seem to do any harm.

owi92 (approved):
Lgtm, thanks!

Arnei:
And thank *you* for thoroughly testing this and even adding all these suggestions to make the code a little nicer :)

PR #1706, opened 2026-05-28
Make editor audio-only proof

### LINKED ISSUE #1636: Support editing of audio-only files
Currently we only render tracks with a video stream in our UIs. Tracks that have only an audio-stream are explicitly ignored.
https://github.com/opencast/opencast/pull/5856 added support for audio-only events to Opencast. Opening the editor for such an event results in half-broken UI. I think the editor frontend should follow suit and allow proper editing of these events.

This would adapting the editor holistically. Each view should be checked separately if it could work with audio-only events, and what changes might be necessary. I'm assuming that the thumbnail and the track view in particular will require additional consideration.

We should also consider if it possible that the editor receives audio files which we do not actually want to display. Maybe because it is common to add them to the same flavor as the video tracks for Opencast workflow purposes, or something along those lines.

Arnei:
Fixes https://github.com/opencast/editor/issues/1636. 

Make sure the editor deals reasonably well with audio-only tracks.

Includes:
- Displaying a little image for the player if there is no video stream.
- Make sure the track selection makes sense (don't display a video stream for an audio only track, don't allow users to deselect all tracks)
- Disable thumbnail generation for audio-only tracks, but keep thumbnail generation.

### How to test this

The default community workflows attempt to circumvent the audio-only issue somewhat by adding dummy video track consisting of one image to the preview tracks for the editor. To generate true audio-only preview tracks for testing, you will want to change the relevant part in `etc/workflows/partial-preview` to something like this:

```yaml
  # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
  # Track previews

  # Create track previews for video editor stream selection feature
  # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
  - id: image
    exception-handler-workflow: partial-error
    description: Create video preview images for presenter and presentation
    configurations:
      - source-flavor: '*/source'
      - target-flavor: '*/video+preview'
      - target-tags: preview
      - encoding-profile: editor.tracks.preview
      - time: 50%

  - id: prepare-av
    if: NOT (${presenter_source_video} OR ${presentation_source_video}) AND ${presenter_source_audio}
    description: Preparing presenter audio and video work versions
    configurations:
      - source-flavor: 'presenter/source'
      - target-flavor: 'presenter/preview'
      - rewrite: false
      - audio-muxing-source-flavors: '*/source'

  - id: prepare-av
    if: NOT (${presenter_source_video} OR ${presentation_source_video}) AND ${presentation_source_audio}
    description: Preparing presenter audio and video work versions
    configurations:
      - source-flavor: 'presentation/source'
      - target-flavor: 'presentation/preview'
      - rewrite: false
      - audio-muxing-source-flavors: '*/source'

  - id: tag
    description: Tagging media package elements
    configurations:
      - source-flavors: 'presenter/preview, presentation/preview'
      - target-tags: preview
      - copy: false
 ```

Make sure the editor still behaves reasonably for normal video files as well.

#### Screenshot:

Bildschirmfoto vom 2026-05-28 15-38-35

### CODE REVIEW

**Review by snoesberger — COMMENTED** [review]
Thank you for improving the editor's ability to handle audio-only files!

My tests were successful, I just have a few minor points to mention:
- In dark mode, the video-off.png image is barely visible: Image
Ideally, the image would behave in the same way as the menu items.
- Once this PR has been merged, the default Opencast workflow should be changed in the Opencast repository as you described.
- It would probably be also a good idea to change the default values for the trackSelection of the editor config `etc/ui-config/mh_default_org/editor/editor-settings.toml`:
```
[trackSelection]

# If the track selection appears in the main menu
# Type: boolean
# Default: true
#show = true

# Ensure that at least one video stream remains selected
# Typically, the track selection ensures that at least one video stream
# remains selected. If you would like your users to be able to create selections
# with only audio streams, set this to false.
# Default: true
atLeastOneVideo = false
```

**Comment by Arnei** [comment]
I am not sure if setting `atLeastOneVideo = false` as the default setting is a good idea. At the very least, I remember other adopters not caring for audio only events and always wanting at least one video selected.

**Comment by snoesberger** [comment]
> I am not sure if setting `atLeastOneVideo = false` as the default setting is a good idea. At the very least, I remember other adopters not caring for audio only events and always wanting at least one video selected.

OK, then. Let's leave it as it is.

**Comment by gregorydlogan** [comment]
The editor and backend work as expected, however Paella 8 barfs with:
```
watch.ts:94 paella-core - Error: Error: Invalid video manifest: the 'streams' array is empty.
(anonymous) @ paella-core.js:5022
(anonymous) @ paella-core.js:5043
error @ paella-core.js:5100
loadManifest @ paella-core.js:6652
await in loadManifest
(anonymous) @ watch.ts:94
```

I'm assuming this is a Paella 8 problem, rather than an editor problem?  I've checked the output file (an `m4a`) is correctly cut, so this appears to be a strictly playback issue.

Arnei:
Proobably a Paella 8 problem? Reminds of the "Note" in this issue https://github.com/opencast/opencast/pull/7547#issue-4216582111

gregorydlogan:
Filed as https://github.com/polimediaupv/paella-opencast/issues/180, since this kinda needs to work.  Merging this since it works.

PR #1712, opened 2026-07-07
Additional GHA bugs

gregorydlogan:
This PR resolves a number of issues related to #1709

Notably, this PR:
- **Adding token, and making built artifacts unique per run**
- **Setting the PR number so that the comment making step can find it.**
- **In some cases, like my test infra, the owner here needs to be templated.**

### DIFF
--- a/.github/workflows/pr-build-test-branch.yml
+++ b/.github/workflows/pr-build-test-branch.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Upload build to GHA
         uses: actions/upload-artifact@v4
         with:
-          name: build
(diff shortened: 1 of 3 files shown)


gregorydlogan:
Note that [this check](https://github.com/opencast/editor/pull/1712/changes#diff-e57c292e6f549d33d6c1349f7cabd7868f73daab0eccb1d734de1a883dda3cc9L77) was failing, and was removed by https://github.com/opencast/editor/pull/1712/changes/daef5d98845b3de10aa6a03b8feae7d28909037d, which is doubly funny since a malicious contributor *would probably remove that step as part of an attack anyway*

PR #1717, opened 2026-08-07
Prevent user from making video too short

Arnei:
Helps with https://github.com/opencast/opencast/issues/5174.

The backend will throw an error during the workflow if all segments in the cutting view are set to "delete", or if the non-deleted segments are all shorter than 2 seconds. This adds a check that prevents a user from sending such an "illegal" segment list to the backend.

### DIFF
--- a/src/i18n/locales/en-US.json
+++ b/src/i18n/locales/en-US.json
@@ -108,7 +108,8 @@
       "startProcessing-button": "Start processing",
       "back-button": "Take me back",
       "selectWF-button": "Click to select this workflow",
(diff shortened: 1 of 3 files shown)


gregorydlogan (changes requested):
Image

Can we remove the 'please wait a bit and try again' when this happens?

Otherwise tested working.

Arnei:
We could, but that particular line is part of the "generic error message" text that has been like that for, *checks git blame*, over 5 years. I agree that the phrasing does not make a ton of sense for this particular error, but I would not want to "just" remove it.

PR #1720, opened 2026-08-20
Redesign editor.opencast.org landing page

lkiesow:
The current design has some overflows for longer branch names like `develop` which makes it look a bit broken. It also leaves aa lot of free and unused space.

More than that, the deployment is very complicated with the main page being included in a huge `echo` command.

The page is now a regular HTML file at `.github/assets/index.html` with a `<!-- VERSIONS -->` marker where the workflow injects the list of deployed versions. Stylesheet and Opencast logo are inlined, so the landing page is a single self-contained file making it much easier to deploy.

The page itself got a fresh look as well: a card with the headline and a short explanation next to a grid of version buttons, a system font stack, dark mode support and a layout which collapses to a single column on narrow screens.

Screenshot from 2026-08-20 23-44-33


Arnei:
Since you are already looking at this, I would like to make a request. Could you make it so the unsupported versions (r/17.x, r/18.x) are dropped from the list? Not only are we not supporting them anymore, but they are probably also using either develop.opencast.org or stable.opencast.org as their backend, which are running different Opencast versions.

gregorydlogan:
Agree with Arne.  Use `.github/get-release-server.sh` in the CI step, if it doesn't say anything then don't add the branch to the generated page.

Also, can you do this to admin and studio too please? :D

PR #1733, opened 2026-09-18
Yet more GHA issues

gregorydlogan:
This PR should fix the GHA script which was failing to correctly deploy the built PR branches due to an incorrect (and unneccessary) check.  This will need to be merged forward to `develop` to take effect, but I'd like to keep all the scripts in sync if we can.


gregorydlogan:
~As an example, look what's working ^^^~

Welp, the comment posts, but the URL and contents are wrong.  So yeah, this will need to get merged all the way forwards once I've corrected that.

